PR #12, opened 2025-03-06
Charge Trapping apps

snpike:
Nuclearizer apps which determine charge trapping correction parameters. One app will take in Am241 illumination data for both sides of the detector, and will separately determine the charge loss parameters on the HV side and the LV side. The other app will take in any dataset and maximize symmetry in HV/LV energy ratio vs. CTD space in order to determine charge loss parameters for one side of the detector while the parameters for the other side remain frozen. The Am241 app will be useful for laboratory calibrations to determine the state of charge trapping before launch, and the symmetry-maximizing app will be useful for ongoing calibration of hole-trapping throughout the mission.


ckierans:
There is an issue with TFitResultPtr when a fit is empty which causes the program to crash (Sean calls ->IsEmpty() at various places, but this is attempting to access a pointer which doesn't exists and it throws an error and seg faults). I will merge this now, but make a new pull request with the necessary changes to fix this bug.

PR #13, opened 2025-03-06
New strip pairing Expos

snpike:
Adding 2 new Expos for strip pairing modules. One plots a 1-D histogram of the number of Hits per Event, and the other plots a 2-D histogram of the number of HV and LV Strip Hits per Hit. Strip pairing modules will also be updated to use these modules.


ckierans:
This PR is now obsolete since #61 has been merged.

PR #15, opened 2025-03-06
Fix for TAC calibration file not remembering.

parshadkp:
Filled in ReadXmlConfiguration and CreateXmlConfiguration.


snpike (changes requested):
We should also save the values of m_MinimumTAC and m_MaximumTAC. MModuleEventFilter can be used as a template for this.

snpike (approved):

PR #16, opened 2025-03-31
No Named Detectors

snpike:
MModuleDepthCalibration2024 now determines detector thicknesses from the geometry file. It also no longer uses Named Detectors. Instead, it looks for detectors with type "Strip3D" which have 1 sensitive volume.

### DIFF
--- a/src/MModuleDepthCalibration2024.cxx
+++ b/src/MModuleDepthCalibration2024.cxx
@@ -111,13 +111,6 @@ MModuleDepthCalibration2024::~MModuleDepthCalibration2024()
 bool MModuleDepthCalibration2024::Initialize()
 {
 
-  if( LoadCoeffsFile(m_CoeffsFile) == false ){
-    return false;
-  }
-  if( LoadSplinesFile(m_SplinesFile) == false ){
-    return false;
-  }
-
   // The detectors need to be in the same order as DetIDs.
   // ie DetID=0 should be the 0th detector in m_Detectors, DetID=1 should the 1st, etc.
   vector<MDDetector*> DetList = m_Geometry->GetDetectorList();
@@ -163,6 +156,14 @@ bool MModuleDepthCalibration2024::Initialize()
     cout<<"No Strip3D detectors were found."<<endl;
     return false; 
   }
+
+  if( LoadCoeffsFile(m_CoeffsFile) == false ){
+    return false;
+  }
+  if( LoadSplinesFile(m_SplinesFile) == false ){
+    return false;
+  }
+
   MSupervisor* S = MSupervisor::GetSupervisor();
   m_EnergyCalibration = (MModuleEnergyCalibrationUniversal*) S->GetAvailableModuleByXmlTag("EnergyCalibrationUniversal");
   if (m_EnergyCalibration == nullptr) {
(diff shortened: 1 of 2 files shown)


src/MModuleDepthCalibration2024.cxx
```diff
@@ -163,6 +156,14 @@ bool MModuleDepthCalibration2024::Initialize()
     cout<<"No Strip3D detectors were found."<<endl;
     return false; 
   }
+
+  if( LoadCoeffsFile(m_CoeffsFile) == false ){
```
zoglauer: Can you apply the normal style conventions here?

src/MModuleDepthCalibration2024.cxx
```diff
@@ -714,8 +715,10 @@ bool MModuleDepthCalibration2024::AddDepthCTD(vector<double> depthvec, vector<ve
 
   double maxdepth = * std::max_element(depthvec.begin(), depthvec.end());
   double mindepth = * std::min_element(depthvec.begin(), depthvec.end());
-  m_Thicknesses[DetID] = maxdepth-mindepth;
-  cout << "MModuleDepthCalibration2024::AddDepthCTD: The thickness of detector " << DetID << " is " << m_Thicknesses[DetID] << endl;
+  if (fabs((maxdepth-mindepth) - m_Thicknesses[DetID]) > 0.0001) {
+    cout<<"MModuleDepthCalibration2024::AddDepthCTD: Warning, the thickness of detector "<<DetID<<" listed in the geometry file does not match the depth-CTD file."<<endl;
+    cout<<"Geometry file gives "<<m_Thicknesses[DetID]<<"cm, while the depth-CTD file gives "<<(maxdepth-mindepth)<<"cm."<<endl;
```
zoglauer, line 725: Maybe return false?

src/MModuleDepthCalibration2024.cxx
```diff
@@ -111,54 +111,59 @@ MModuleDepthCalibration2024::~MModuleDepthCalibration2024()
...
+          m_NYStrips[DetID] = strip->GetNStripsY();
+          cout << "Found detector " << det_name << " corresponding to DetID=" << DetID << "." << endl;
+          cout << "Detector thickness: " << m_Thicknesses[DetID] << endl;
+          cout << "Number of X strips: " << m_NXStrips[DetID] << endl;
+          cout << "Number of Y strips: " << m_NYStrips[DetID] << endl;
+          cout << "X strip pitch: " << m_XPitches[DetID] << endl;
+          cout << "Y strip pitch: " << m_YPitches[DetID] << endl;
+          m_DetectorIDs.push_back(DetID);
+          m_Detectors[DetID] = det;
+        }
```
zoglauer, line 152: error message here

src/MModuleDepthCalibration2024.cxx
```diff
@@ -111,54 +111,59 @@ MModuleDepthCalibration2024::~MModuleDepthCalibration2024()
...
+      }
```
zoglauer, line 155: Error message here

zoglauer:
Looks good. Just try to add error messages if something fails, so that it is easier to debug.

src/MModuleDepthCalibration2024.cxx
snpike: Do you think I should return false in these cases or let the user know but allow the module to continue running?

zoglauer:
Are you OK with me merging it?

snpike:
> Are you OK with me merging it?

yes please

PR #18, opened 2025-05-02
Nanosecond tac calibration

snpike:
This pull request updates the TAC calibration module to perform a more sophisticated TAC cut which takes into account information regarding the coincidence window, shaping offset, disable time, and flag to enable delay. The latter two quantities are assumed to be constant across the entire dataset (i.e. they are not dependent upon strip, ASIC, or detector). The values of the coincidence window and shaping offset must be provided for each strip via an input file.

Additional changes:
- energy calibration now occurs prior to TAC calibration
- The TAC Expo plots the calibrated value rather than the raw TAC, and only plots timing values which were not cut


zoglauer:
Looks good, merge it.

PR #20, opened 2025-05-22
CHG: Calculate X and Y position using energy-weighted strip positions

snpike:
Rather than just placing Hits directly in the middle of the strips with the most energy, now the X- and Y-positions are calculated using charge sharing information by weighting each strip by the amount of energy deposited.

### DIFF
--- a/include/MGUIOptionsDepthCalibration2024.h
+++ b/include/MGUIOptionsDepthCalibration2024.h
@@ -77,6 +77,8 @@ class MGUIOptionsDepthCalibration2024 : public MGUIOptions
   //! Check button if working with the Card Cage at UCSD
   TGCheckButton* m_UCSDOverride;
 
+  //! Check button if weighting X and Y by energy
+  TGCheckButton* m_WeightedXY;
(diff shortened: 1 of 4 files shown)


zoglauer:
Can you make that optional in the UI?
Since the most energy deposit of an electron track is at its end where the Bragg peak is, it is not immediately ensured that this weighting by energy deposit will always give the best locations, especially at higher energies.

ckierans:
Discussed in CDEE Meeting on Sept 23:
As Andreas requests above, we implement this, but should make it optional. We hope to develop other methods to determine the x/y position based on the strips hit and the energies (ML, anyone?), so we will likely have a few options down the line.

We also need to think about an approach to test this.

ckierans:
I checked this out quickly and I'm getting a slightly large ARM when using the weighted positions (8.7 deg for HP52358-3 gse_20250709T113820.hdf) vs not using the weighted positions (8.4 deg). This will require a closer look, perhaps after the coordinate system has been confirmed.

PR #24, opened 2025-06-03
New flags for Strip Hits

snpike:
Added 3 new boolean flags indicating whether a Strip Hit has good timing that can be used for calibration, whether it is a guard ring hit, and whether it is a nearest neighbor hit. Corresponding functions to turn the nearest neighbor and guard ring flags into an unsigned integer to be printed into data files.


zoglauer:
OK merge it. We have to do some modifications in melinator to read the new roa file, but it can do that. But I can do that.

PR #25, opened 2025-06-10
Fast timing and calibration flags

snpike:
Adding a fast timing flag to MStripHit and changed "good" timing flag to calibrated timing flag which indicates whether the TAC -> ns timing calibration has been completed for a Strip Hit.


snpike:
Wrong branch, sorry

PR #36, opened 2025-09-23
Depth splines

snpike:
Updated parsing and handling of depth-CTD splines in order to sample the full depth of the detector and to ensure that the depth vector is evenly spaced. This pull request also includes some comment cleanup and minor bug fixes.


ckierans (commented):
Ran through Sean's updated depth code and confirmed the ARM distribution is preserved. 

The hit distribution is still not quite right (see below) and these are captured in current issues: 1) The bump will persist until we handle the cross-talk between nearest neighbors, 2) we already have an issue to clean up the hits near the face of each detector. This should come from a better modeling of the depth noise. 

Screenshot 2025-12-09 at 9 53 10 AM

PR #60, opened 2025-10-24
Updated Energy Calibration module with option for soft threshold, inc…

### LINKED ISSUE #39: Implement an energy/ADC cut slightly above the slow threshold
There is a noise peak below the slow threshold for triggered strips that needs to be cut in the Energy Calibration Module.
See slide 18-19 of Clio’s ASIC slides from COSIFest-2024

Image

Emily-Broadbent:
This PR closes #39 

I have updated the energy calibration module to include a soft energy threshold (keV). I removed the previous Temperature options, and there are now two new options in the nuclearizer GUI. The first option is to include a soft threshold calibration file, which should have a column of threshold values unique to each strip. When this is enabled, the module will remove the SH if the energy is below the threshold, if there is no threshold given a default threshold of 15 keV will be applied. The second option is for one user-input threshold that will be applied to all strips. 
Screenshot 2025-10-24 at 10 27 31 AM
Screenshot 2025-10-24 at 10 27 51 AM

Example using the threshold file:
Screenshot 2025-10-24 at 10 29 06 AM

Example using a user-input threshold of 50 keV, to easily see the cut:
Screenshot 2025-10-24 at 10 30 20 AM


zoglauer:
Hi Emily,

m_ThresholdFileEnabled & m_UseThresholdValue are booleans, thus don't use 0 and 1 for comparison, but true and false

volatge -> voltage

NULL is for C pointers, what's the default value here if it is not defined? Likely 0
double Threshold_map = m_ThresholdMap[R]; // if file enabled, declare value from map
if (Threshold_map == NULL
 
Remove your //raise flag - it's just an error message printed to the screen

All TODO's should be on a new line, and an issue should be added

zoglauer:
Looks OK. 

One minor modification: Put your, Carolyn's and Jacqueline Beechert's name in the copyright lane

As soon as Carolyn gives the OK. You can merge it in.

ckierans:
Looks good, Emily! Nice work. Andreas can make requested GUI changes after this merge.

PR #67, opened 2025-10-27
DEE Shield energy correction - loading the file

vfioretti:
The DEE Shield Energy Correction sub module is modified to configure and read the correction file. The code still does nothing to it.

MModuleDEESMEX and MGUIOptionsDEESMEX files also modified to modify the GUI options.

Code checked with clang-factor with this line commented: InsertBraces: true


zoglauer:
Looks good.

fhagemann:
I'm getting an error now if I don't pass any Shield correction DEE file. 
Should this file be required or can it be optional?
If it should always be passed: where can I find a dummy file, or can I just pass it any CSV file I like?
image

PR #68, opened 2025-10-28
Ability to read calibrated energy values from ROA files

fhagemann:
I shamelessly copied the code from https://github.com/cositools/nuclearizer/commit/876ef0dd3e5e23ecfa0900c028ab00ef9c1202b5 (to read in TAC values from ROA files) and used it to write a parser for calibrated energy, which is saved using the identifier `energy` and then parsed as `double`.

To do a quick cross-check that it works, I ran this (reading an ROA file and writing another ROA file again)
image
and compared the results

Left is the old file and right is the new file
image

Energies seem to be parsed correctly and written to the file, yielding the same results.
Only difference right now is that the values for `CL` (?) are lost, but I'll leave this to a future PR.

### DIFF
--- a/include/MReadOutDataEnergy.h
+++ b/include/MReadOutDataEnergy.h
@@ -1,0 +1,111 @@
+/*
+ * MReadOutDataEnergy.h
+ *
+ * Copyright (C) by Andreas Zoglauer.
+ * All rights reserved.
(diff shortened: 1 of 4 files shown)


zoglauer:
Approved.

PR #70, opened 2025-10-28
[DEE] Update dummy implementation of strip IDs

fhagemann:
I was getting warnings in nuclearizer when applying a realistic Ecal file to DEE output, 
saying that no energy-fit was found for HV and LV strip 65.

To my knowledge, we use strip IDs 0-63 for the strips and 64 for the guard ring (at least that's how I read the ecal file),
so I updated the implementation of this to return strip IDs from 0 to 63 for the strips and 64 for the GR.

After updating this, I was able to simulate my first pixel/strip maps for an Am241 source placed at the same distance that we use in lab measurements :tada:

image
image
image

### DIFF
--- a/src/MSubModuleChargeTransport.cxx
+++ b/src/MSubModuleChargeTransport.cxx
@@ -98,11 +98,11 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
   for (MDEEStripHit& SH: LVHits) {
     MVector Pos = SH.m_SimulatedPositionInDetector;
     int ID = (Pos.X() + 7.4/2) / (7.4/64);
-    if (ID > 0 && ID < 64) {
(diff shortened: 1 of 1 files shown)


zoglauer:
Cool. Why do you need the "&& Pos.X() >= -7.4/2"? Isn't that covered in ID >= 0?

fhagemann:
If you do integer division, you round towards zero. If I don’t do it, I end up getting a strip 0 which is double the width (extending one strip with towards negative x/y).

Maybe, there is a more elegant solution. This is the quick fix I came up with for now.

zoglauer:
Ahh truncation towards zero - forgot about that. Can you try:
```
// Calculate strip ID by rounding down intentionally to avoid truncation towards zero

int ID = static_cast<int>(std::floor((Pos.X() + 7.4/2.0) / (7.4/64.0)));
```

PR #73, opened 2025-11-22
[DEE] Improve strip ID handling for guard ring events

fhagemann:
In the DEE, the way the strip ID is determined is ignoring the shape of the GR.

Current implementation:
image

After this PR:
image

The exact values I put here are still dummy values, which should be cross-checked with the SMEX massmodel in the future.


ckierans (commented):
Can you add a line like
"// TODO: Confirm the correct boundary of the guard ring based on SMEX detector models"
to get make sure this is documented in the code?

There is also a related issue to include the mask metrology measurements into the DEE. This isn't a high priority relative to other more major DEE implementation steps.

ckierans (approved):
Looks good!

PR #75, opened 2025-12-02
Slow threshold cut UI updates

### LINKED ISSUE #65: allow only one input option in MGUIOptionsEnergyCalibrationUniversal
Currently, in the MGUIOptionsEnergyCalibrationUniversal file, it allows both the ThresholdValue checkbox and ThresholdFile check box to be applied at the same time. Should be changed to allow one or the other, not both

zoglauer:
I modified the UI to what I think it was intended to look like:
+ I replaced the check buttons with radio buttons
+ cleaned up the code
+ I tried to keep the core functionality untouched, but have no slow threshold cut file to test it

Thus, Emily, could you take a look and test it thoroughly?

### DIFF
--- a/include/MGUIOptionsEnergyCalibrationUniversal.h
+++ b/include/MGUIOptionsEnergyCalibrationUniversal.h
@@ -43,6 +43,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 
+//! The user interface for the universal energy calibration
 class MGUIOptionsEnergyCalibrationUniversal : public MGUIOptions
 {
   // public Session:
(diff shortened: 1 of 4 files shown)


zoglauer:
PS: The bug that the file name text can spill out of the text box is a ROOT bug which I have reported.
https://github.com/root-project/root/issues/20602

zoglauer:
Reminder: Close https://github.com/cositools/nuclearizer/issues/65 when this is merged

fhagemann:
Does this fix that the energy threshold was fixed to 35keV ?

Emily-Broadbent:
This should solve the fixed threshold, but I am still learning git and probably merged when I was not supposed to.  I was actually trying to figure out how to sync the previous commit to my own local code, I tried all of the git commands I know and either Andreas's code did not solve the GUI input, or I never was able to sync properly. Sorry for any confusion

Emily-Broadbent:
I have checked @zoglauer 's code and it allows for the requested GUI functions and properly applies threshold cuts for a set value and single input value.

PR #76, opened 2025-12-04
[DEE] Read detector dimensions from the geometry

fhagemann:
Changing `MSubModuleChargeTransport` all at once became kind of chaotic, so I will be splitting my PR into smaller PRs, which should make it way easier to review and to collect feedback for the individual steps.

In this PR, I'm replacing the hard-coded values for strip pitches, detector width, number of strips etc. by the numbers obtained from `MModule::m_Geometry`, which I had to pass down to `MSubModuleChargeTransport`.

I will mark the parts of the codes where I have questions/comments after opening the PR.

### DIFF
--- a/src/MSubModuleChargeTransport.cxx
+++ b/src/MSubModuleChargeTransport.cxx
@@ -32,6 +32,8 @@
 
 // MEGAlib libs:
 #include "MSubModule.h"
+#include "MDShapeIntersection.h"
+#include "MDShapeTUBS.h"
 
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -93,10 +95,23 @@ bool MSubModuleChargeTransport::Initialize()
           m_NYStrips[DetID] = strip->GetNStripsY();
           m_XWidths[DetID] = strip->GetWidthX();
           m_YWidths[DetID] = strip->GetWidthY();
+
+          // Read the detector radius from the geometry (assuming it is the second shape of an intersection)
+          if (vol->GetShape()->GetType() == "Intersection" && dynamic_cast<MDShapeIntersection*>(vol->GetShape())->GetShapeB()->GetType() == "TUBS") {
+            m_Radii[DetID] = dynamic_cast<MDShapeTUBS*>(dynamic_cast<MDShapeIntersection*>(vol->GetShape())->GetShapeB())->GetRmax();
+          } else {
+            // If that does not exist, set the detector radius to a value high enough to not have an impact
+            m_Radii[DetID] = m_XWidths[DetID] + m_YWidths[DetID];
+            if (g_Verbosity >= c_Info) {
+              cout << m_Name << ": No bounding tube volume found for this detector" << endl;
+            }
+          }
+          
           if (g_Verbosity >= c_Info) {
             cout << "Found detector " << det_name << " corresponding to DetID=" << DetID << "." << endl;
             cout << "Detector width (X): " << m_XWidths[DetID] << endl;
             cout << "Detector width (Y): " << m_YWidths[DetID] << endl;
+            cout << "Detector radius (R): " << m_Radii[DetID] << endl;
             cout << "Detector thickness: " << m_Thicknesses[DetID] << endl;
             cout << "Number of X strips: " << m_NXStrips[DetID] << endl;
             cout << "Number of Y strips: " << m_NYStrips[DetID] << endl;
@@ -153,6 +168,7 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     double XWidth = m_XWidths[DetID];
     double YWidth = m_YWidths[DetID];
     double XPitch = m_XPitches[DetID];
+    double Radius = m_Radii[DetID];
     int NXStrips = m_NXStrips[DetID];
 
     // Calculate LV strip ID by rounding down intentionally to avoid truncation towards zero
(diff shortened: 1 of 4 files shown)


src/MSubModuleChargeTransport.cxx
```diff
@@ -98,17 +148,23 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+
     // Calculate LV strip ID by rounding down intentionally to avoid truncation towards zero
-    // TODO: Confirm the correct strip pitch based on SMEX detector models
-    int ID = static_cast<int>(std::floor((Pos.X() + 7.4/2.0) / (7.4/64.0)));
+    int ID = static_cast<int>(std::floor((Pos.X() + XWidth/2.0) / XPitch));
 
     // Check for strip ID and if the position is within the allowed strip length or on the guard ring
     // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
-    if (ID >= 0 && ID < 64 && std::abs(Pos.Y()) <= 7.4/2.0 && std::hypot(Pos.X(), Pos.Y()) <= 4.712) {
+    if (ID >= 0 && ID < NXStrips && std::abs(Pos.Y()) <= YWidth/2.0 && std::hypot(Pos.X(), Pos.Y()) <= 4.712) {
```
fhagemann: The only number that is not obvious for me to find is the radius of the guard ring for the edge strips, so that's why in here it's still `4.712`. Is that number stored somewhere in the geometry?

src/MSubModuleChargeTransport.cxx
```diff
@@ -118,17 +174,24 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+
     // Calculate HV strip ID by rounding down intentionally to avoid truncation towards zero
     // TODO: Confirm the correct strip pitch based on SMEX detector models
-    int ID = static_cast<int>(std::floor((Pos.Y() + 7.4/2.0) / (7.4/64.0)));
+    int ID = static_cast<int>(std::floor((Pos.Y() + YWidth/2.0) / YPitch));
 
     // Check for strip ID and if the position is within the allowed strip length or on the guard ring
     // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
-    if (ID >= 0 && ID < 64 && std::abs(Pos.X()) <= 7.4/2.0 && std::hypot(Pos.X(), Pos.Y()) <= 4.712) {
+    if (ID >= 0 && ID < NYStrips && std::abs(Pos.X()) <= XWidth/2.0 && std::hypot(Pos.X(), Pos.Y()) <= 4.712) {
```
fhagemann: Same thing here.

src/MSubModuleChargeTransport.cxx
```diff
@@ -98,17 +148,23 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
 
     // Check for strip ID and if the position is within the allowed strip length or on the guard ring
     // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
-    if (ID >= 0 && ID < 64 && std::abs(Pos.Y()) <= 7.4/2.0 && std::hypot(Pos.X(), Pos.Y()) <= 4.712) {
+    if (ID >= 0 && ID < NXStrips && std::abs(Pos.Y()) <= YWidth/2.0 && std::hypot(Pos.X(), Pos.Y()) <= 4.712) {
       SH.m_ROE.SetStripID(ID);
       SH.m_IsGuardRing = false;
     } else {
-      SH.m_ROE.SetStripID(64);
+      SH.m_ROE.SetStripID(NXStrips);
```
fhagemann, line 183: Is there any objections setting the GR strip ID to the number of strips on that side?
For COSI SMEX, we have 64 strips with IDs 0 to 63, and we give the GR an ID of 64, just like in the data.

src/MSubModuleChargeTransport.cxx
zoglauer: Hmm, I have an IngotRadius of 2 x 2.54 (2 inch) that minus the guard ring size is 4.78?
Where did you get your number?

Try:
MDShapeTUBS* Shape = dynamic_cast<MDShapeTUBS*>(Geometry.GetShape("WaferCutDisk"));
if (Shape != nullptr) {
   cout<<"Ingot radius: "<<Shape.GetRmax()<<endl;
}

src/MSubModuleChargeTransport.cxx
zoglauer: see above

src/MSubModuleChargeTransport.cxx
zoglauer: Should be the same output as the hard ware - can you ask Alex or Brent?

src/MSubModuleChargeTransport.cxx
fhagemann: I calculated the number from the strip length values that I saw in a Slack message. But I will look into getting them from the massmodel using the `WaferCutDisk` as suggested.

src/MSubModuleChargeTransport.cxx
fhagemann: Yes, GR has ID 64 on both HV and LV side.

src/MSubModuleChargeTransport.cxx
fhagemann: Ok, this works for files with exactly one detector, but it might not work for geometries with multiple detectors.
I would try to read the `TUBS` from the `SensitiveVolume` (should be `ShapeB` from the `Intersection`), and extract the radius from there.

zoglauer:
I think this is OK for now.
We have to come up with a plan how to handle, especially name, all the 16 detectors in their holders. In the past we had a file with all the parameters, and had a python script to generate our 12 detectors. Maybe think if we should do something similar.

fhagemann:
Can we merge this PR or should I add a TODO somewhere first and/or open an issue about this?

zoglauer:
It's this:
https://github.com/cositools/nuclearizer/issues/77

PR #79, opened 2025-12-10
ADD: Module to perform revan reconstruction

zoglauer:
First version of an event reconstruction module.
Please take a look. Feedback welcome.


julianmgerber (approved):
I tested the Revan module on 361-1 data to see if I could reproduce the ARM. Everything ran successfully and I was able to get an ARM of 8.16 degrees (which is the same as the FWHM I was getting previously).

Issue #50, opened 2025-10-01
Check all BD flags!


ckierans:
Clean up the BD flags from the balloon days, and update the tracking of these for SMEX

ckierans:
We may need another flag for Quality of events. For example, we'd want to track when we've removed strip hits below the slow software threshold, but we don't want these events to be removed through a BD flag. Same as the chi square cut in the strip pairing.

PR #81, opened 2025-12-18
Updated strip hit threshold logic

ckierans:
Julian reported that the slow threshold cut was dramatically decreasing the number of events in the final ARM plots (using cut of 35 keV). 

I confirmed with HP52358-3 gse_20250709T113820.hdf that I recovered only 30% of Compton events in revan when the slow threshold of 35 keV was applied.

The changes implemented:
- Updated the logic of the threshold loop so the other strip hits in the event are calibrated ("return false" meant the whole event was skipped if one SH was found to have below threshold energy)
- Updated the error code so that a c_Warning is created for any SH below threshold, and the energy is included in the warning message
- Updated the ReadAssembly so the events with a SH below threshold now get a "QA" flag instead of a "BD" flag so these events aren't filtered out in revan. (As part of the other issue to go through all BD flags in Nuclearizer (#50), some other flags will use this QA nomenclature)
- ran clang-format on MModuleEnergyCalibrationUniversal so there are lots of formatting fixes

After these changes we now recover 71% of Compton events in revan with the 35 keV threshold (and 99.5% of the Compton events when using a slow threshold file with realistic thresholds ~10 keV). I confirmed the ARM isn't dramatically altered.


src/MModuleEnergyCalibrationUniversal.cxx
```diff
@@ -296,110 +305,134 @@ bool MModuleEnergyCalibrationUniversal::Initialize()
...
       }
-      
-      //! Remove SH for any energy value below the established threshold
+
+      //! Remove SH for any energy value below the established threshold (0 is default)
       if (Energy < Threshold) {
+        if (g_Verbosity >= c_Warning) {
+          cout << m_XmlTag << ": Strip Hit below threshold, deleting SH with Energy " << Energy << " keV " << endl;
+        }
+        Event->SetStripHitBelowThreshold(true, "with Energy " + to_string(Energy));
```
zoglauer: if we have multiples this gets overwritten. Do we want that?
Or do we want to keep the smallest value?

src/MModuleEnergyCalibrationUniversal.cxx
```diff
@@ -296,110 +305,134 @@ bool MModuleEnergyCalibrationUniversal::Initialize()
...
-      double Threshold = 0; // 0 means ignore
+
+      if (Energy < 0) {
+        Energy = 0; // If the calibrated energy is less than 0, force it to be 0.
+      }
       
       if (m_SlowThresholdCutMode == MSlowThresholdCutModes::e_Fixed) { //check if user input threshold is enabled (one value applied to all strips)
         Threshold = m_SlowThresholdCutFixedValue;
       } else if (m_SlowThresholdCutMode == MSlowThresholdCutModes::e_File) { //check if threshold file is enabled (unique value applied to each strip)
         double Threshold_map = m_ThresholdMap[R]; // if file enabled, declare value from map
```
zoglauer: Rename to ThresholdFromFile

src/MModuleEnergyCalibrationUniversal.cxx
```diff
@@ -296,110 +305,134 @@ bool MModuleEnergyCalibrationUniversal::Initialize()
...
-        
+
         if (Threshold_map == 0) {
-          if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Error: Threshold not found for read-out element "<<R<<endl;
-          Threshold = 15;   // set default threshold if threshold not found
+          if (g_Verbosity >= c_Error) {
+            cout << m_XmlTag << ": Error: Threshold not found for read-out element " << R << endl;
+          }
+          Threshold = 15; // set default threshold if threshold not found
```
zoglauer: It is better to define "magic numbers" explicitly in the beginning:

const double DefaultSlowThreshold = 15.0;

zoglauer: Think about if this might actually be part of the map file ?

include/MReadOutAssembly.h
```diff
@@ -451,10 +450,12 @@ class MReadOutAssembly : public MReadOutSequence
   MString m_DepthCalibrationIncompleteString;
   bool m_DepthCalibration_OutofRange;
   MString m_DepthCalibration_OutofRangeString;
-  bool m_StripHitBelowThreshold;
-  MString m_StripHitBelowThresholdString;
 
 
+  // Flags indicating the quality of the event: quality warning, but not to be filtered out
```
zoglauer: Although the names are already long:
...Incomplete always results in a bad flag
... Do you want something similar for the quality flags?

Think of soemthing - or something similar
m_StripHitBelowThresholdQF;
m_StripHitBelowThresholdQualityFlag;

zoglauer:
I didn't check if it runs, I just reviewed the code.

src/MModuleEnergyCalibrationUniversal.cxx
ckierans: Good catch. Maybe instead we should track how many SH we loose and all of the energies...? 
I can edit that.

zoglauer: Maybe:
AddStripHitBelowThreshold(Energy)
And add a bit of logic into the event to track it?
How many hits, and total loss seems reasonable.

ckierans:
@zoglauer, I've address all of your comments, if you want to take another look. I'll open up another PR that cleans up the rest of the BD and QA flags.

Emily-Broadbent (commented):
After @zoglauer 's updates, the logic looks good to me

Issue #69, opened 2025-10-28
Clean up time handling


zoglauer:
Currently we still have COSI-balloon's CL - clock - values in the code.
However, from the space craft we will get GPS time, thus TI is OK - not sure what we get from the ground calibration, probably also UTC?

ckierans:
Discussed plans forward in slack.

The event timestamp in the L1a data format will be “Mission time in seconds since Jan 1, 2025” - this is the “Reference Time System” defined in the COSI HEASARC Tech Agreement. (FYI, 2025 was chosen since that’s the earliest date we might have for calibration data in HEASARC.) There will be a conversion from L0->L1a to handle the RTB truncated time + ICP time + whatever other calculations need to be done to determine the actual event time (I assume Wing is going to handle this?).

Need to define a new variable TimeRTS (i.e., GetTimeRTS and SetTimeRTS) as a double precision floating point number to define the number of seconds since the Jan 1 2025 reference time system (in TT, not UTC). RTS will only be used within Nuclearizer when reading in L1a FITS files that contain this info. 

We still will stream the UTC timestamps to “TI” in the ROA and EVTA file types, as it’s currently being done. 

We will remove the internal “GetTI/SetTI” variable to avoid confusion. 
We will remove the CL variable in the ROA/EVTA files and within Nuclearizer.

Still TBD if we should also stream the RTS data to roa/evta files.

ckierans:
This issue was resolved with PR #124

PR #85, opened 2026-01-13
[DEE] Update `AnalyzeEvent` in `MSubModuleChargeTransport`

fhagemann:
This is a first PR to replace the dummy code in  `MSubModuleChargeTransport::AnalyzeEvent`.
I will open this as a draft PR, there are still a lot of magic numbers in there (bias voltage, temperature, initial charge cloud size, cross-talk parameters), but this is mainly to collect feedback on my current implementation while moving forward and improving this in comparison with data.

Some mathematical background:
The linear charge density of an event at a given position and depth can be approximated by Eq. (7) in https://doi.org/10.1016/j.nima.2023.168310. To get to the actual charge/energy, we need to integrate the linear charge density from the left edge of the strip to the right edge of the strip:
image

The idea is then to calculate the energy for the main strip and it's direct neighbors.
Here is a plot of what is happening in the code:

image

The implementation of cross-talk is based on the parameterization that we obtained from simulations (for now just for one detector and implemented as magic numbers). This currently seems to underestimate the cross-talk on the HV side and overestimate the cross-talk on the LV side.
image

### DIFF
--- a/src/MSubModuleChargeTransport.cxx
+++ b/src/MSubModuleChargeTransport.cxx
@@ -27,8 +27,10 @@
 #include "MSubModuleChargeTransport.h"
 
 // Standard libs:
+#include <cmath>
 
 // ROOT libs:
+#include "TMath.h"
 
 // MEGAlib libs:
 #include "MSubModule.h"
@@ -156,9 +158,23 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
 {
   // Main data analysis routine, which updates the event to a new level 
 
-  // Dummy code:
+  // Define physical constants
+  constexpr double kB = TMath::K(); // unit: J/K
+  constexpr double ElementaryCharge = TMath::Qe(); // unit: C
+  constexpr double IonizationEnergy = 0.00295; // unit: keV
+  constexpr double Epsilon0 = 8.85418781762039e-14; // unit: F/cm
+  constexpr double EpsilonR = 16.0; // in germanium, unitless
+
+  // TODO: Read bias voltage and temperature of the detector from a file
+  constexpr double BiasVoltage = 600.0; // unit: V
+  constexpr double Temperature = 87.0; // unit: K
+
+  // TODO: Implement energy-dependent initial charge-cloud sizes
+  constexpr double InitialChargeCloudSize = 0.02; // 200µm in cm
 
   // Create strip hits
+  list<MDEEStripHit> ChargeTransportLVHits;
+
   list<MDEEStripHit>& LVHits = Event->GetDEEStripHitLVListReference();
   for (MDEEStripHit& SH: LVHits) {
     MVector Pos = SH.m_SimulatedPositionInDetector;
@@ -167,9 +183,13 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     unsigned int DetID = SH.m_ROE.GetDetectorID();
     double XWidth = m_XWidths[DetID];
     double YWidth = m_YWidths[DetID];
+    double Thickness = m_Thicknesses[DetID];
     double XPitch = m_XPitches[DetID];
     double Radius = m_Radii[DetID];
     int NXStrips = m_NXStrips[DetID];
+
+    double MeanElectricField = BiasVoltage / Thickness; // unit: V/cm
+    double N = SH.m_SimulatedEnergy / IonizationEnergy;
 
     // Calculate LV strip ID by rounding down intentionally to avoid truncation towards zero
     int ID = static_cast<int>(std::floor((Pos.X() + XWidth/2.0) / XPitch));
@@ -177,14 +197,86 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     // Check for strip ID and if the position is within the allowed strip length or on the guard ring
     // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
     if (ID >= 0 && ID < NXStrips && std::abs(Pos.Y()) <= YWidth/2.0 && std::hypot(Pos.X(), Pos.Y()) <= Radius) {
-      SH.m_ROE.SetStripID(ID);
-      SH.m_IsGuardRing = false;
+
+      // Apply charge sharing based on relative coordinate to the gap of that strip (0 <= X < XPitch)
+      double XFromGap = std::fmod(Pos.X() + XWidth/2.0, XPitch);
+
+      // calculate σ and µ, assuming t = z / v = z / (µ * E)
+      double SigmaX = std::sqrt(2.0 * kB * Temperature * (Pos.Z() + Thickness / 2.0) / (ElementaryCharge * MeanElectricField)); // in cm
+      double EtaX   = std::cbrt(std::pow(InitialChargeCloudSize, 3) + 3.0 * N * ElementaryCharge * (Pos.Z() + Thickness / 2.0) / (4.0 * TMath::Pi() * Epsilon0 * EpsilonR * MeanElectricField)); // in cm
+      cout << "X: " << Pos.X() << ", Y: " << Pos.Y() << ", Z: " << Pos.Z() << ", E: " << SH.m_SimulatedEnergy << endl;
+      cout << "Sigma X: " << SigmaX << ", EtaX: " << EtaX << endl;
+
+      auto Lambda = [&](double x) -> double {
+        double a = (x - EtaX) / (TMath::Sqrt2() * SigmaX);
+        double b = (x + EtaX) / (TMath::Sqrt2() * SigmaX);
+        return SH.m_SimulatedEnergy / (8.0 * std::pow(EtaX, 3)) * (
+          std::erf(b) * (2.0 * std::pow(EtaX, 3) + x * (3.0 * std::pow(EtaX, 2) - 3.0 * std::pow(SigmaX, 2) - std::pow(x, 2))) + 
+          std::erf(a) * (2.0 * std::pow(EtaX, 3) - x * (3.0 * std::pow(EtaX, 2) - 3.0 * std::pow(SigmaX, 2) - std::pow(x, 2))) + 
+          std::exp(- std::pow(b,2)) * std::sqrt(2.0 / TMath::Pi()) * SigmaX * (EtaX * x + (2.0 * std::pow(EtaX, 2) - 2.0 * std::pow(SigmaX, 2) - std::pow(x, 2))) + 
+          std::exp(- std::pow(a,2)) * std::sqrt(2.0 / TMath::Pi()) * SigmaX * (EtaX * x - (2.0 * std::pow(EtaX, 2) - 2.0 * std::pow(SigmaX, 2) - std::pow(x, 2))) 
+        );
+      };
+
+      double MainStripEnergy    = Lambda(XPitch - XFromGap) - Lambda(-XFromGap);
+      double NNLeftStripEnergy  = Lambda(-XFromGap) - Lambda(-XPitch - XFromGap);
+      double NNRightStripEnergy = Lambda(2.0*XPitch - XFromGap) - Lambda(XPitch - XFromGap);
+
+      cout << "Energy: " << SH.m_SimulatedEnergy << ", split into " << NNLeftStripEnergy << ", " << MainStripEnergy << " and " << NNRightStripEnergy << endl;
+
+      // create entry for the main hit
+      MDEEStripHit MainSH = SH;
+      MainSH.m_ROE.SetStripID(ID);
+      MainSH.m_Energy = MainStripEnergy;
+      MainSH.m_IsGuardRing = false;
+      ChargeTransportLVHits.push_back(MainSH);
+
+      // create MDEEStripHit for the left NN
+      if (NNLeftStripEnergy > IonizationEnergy) {
+        MDEEStripHit NNLeftSH = SH;
+        NNLeftSH.m_Energy = NNLeftStripEnergy;
+        if (ID > 0) {
+          NNLeftSH.m_ROE.SetStripID(ID - 1);
+          NNLeftSH.m_IsGuardRing = false;
+          // NNLeftSH.m_IsNearestNeighbor = true;
+        } else {
+          NNLeftSH.m_ROE.SetStripID(NXStrips);
+          NNLeftSH.m_IsGuardRing = true;
+        }
+        ChargeTransportLVHits.push_back(NNLeftSH);
+      }
+      
+      // create MDEEStripHit for the right NN
+      if (NNRightStripEnergy > IonizationEnergy) {
+        MDEEStripHit NNRightSH = SH;
+        NNRightSH.m_Energy = NNRightStripEnergy;
+        if (ID < NXStrips - 1) {
+          NNRightSH.m_ROE.SetStripID(ID + 1);
+          NNRightSH.m_IsGuardRing = false;
+          // NNRightSH.m_IsNearestNeighbor = true;
+        } else {
+          NNRightSH.m_ROE.SetStripID(NXStrips);
+          NNRightSH.m_IsGuardRing = true;
+        }
+        ChargeTransportLVHits.push_back(NNRightSH);
+      }
+
     } else {
+      // TODO: implement charge sharing also for GR events
+      SH.m_Energy = SH.m_SimulatedEnergy;
       SH.m_ROE.SetStripID(NXStrips);
       SH.m_IsGuardRing = true;
+      ChargeTransportLVHits.push_back(SH);
     }
-    SH.m_Energy = SH.m_SimulatedEnergy;
   }
+
+  // replace old list by new list
+  Event->GetDEEStripHitLVListReference().clear();
+  for (MDEEStripHit& SH: ChargeTransportLVHits) {
+    Event->AddDEEStripHitLV(SH);
+  }
+
+  list<MDEEStripHit> ChargeTransportHVHits;
 
   list<MDEEStripHit>& HVHits = Event->GetDEEStripHitHVListReference();
   for (MDEEStripHit& SH: HVHits) {
@@ -194,9 +286,13 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     unsigned int DetID = SH.m_ROE.GetDetectorID();
     double XWidth = m_XWidths[DetID];
     double YWidth = m_YWidths[DetID];
+    double Thickness = m_Thicknesses[DetID];
     double YPitch = m_YPitches[DetID];
     double Radius = m_Radii[DetID];
     int NYStrips = m_NYStrips[DetID];
+
+    double MeanElectricField = BiasVoltage / Thickness; // unit: V/cm
+    double N = SH.m_SimulatedEnergy / IonizationEnergy;
 
     // Calculate HV strip ID by rounding down intentionally to avoid truncation towards zero
     // TODO: Confirm the correct strip pitch based on SMEX detector models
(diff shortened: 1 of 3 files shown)


src/MSubModuleChargeTransport.cxx
```diff
@@ -156,9 +158,23 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
-  // Dummy code:
+  // Define physical constants
+  constexpr double kB = TMath::K(); // unit: J/K
+  constexpr double ElementaryCharge = TMath::Qe(); // unit: C
+  constexpr double IonizationEnergy = 0.00295; // unit: keV
+  constexpr double Epsilon0 = 8.85418781762039e-14; // unit: F/cm
+  constexpr double EpsilonR = 16.0; // in germanium, unitless
+
+  // TODO: Read bias voltage and temperature of the detector from a file
+  constexpr double BiasVoltage = 1050.0; // unit: V
```
fhagemann: Do we store the bias voltages somewhere for these detectors?

src/MSubModuleChargeTransport.cxx
```diff
@@ -167,25 +183,108 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     unsigned int DetID = SH.m_ROE.GetDetectorID();
     double XWidth = m_XWidths[DetID];
     double YWidth = m_YWidths[DetID];
+    double Thickness = m_Thicknesses[DetID];
     double XPitch = m_XPitches[DetID];
     double Radius = m_Radii[DetID];
     int NXStrips = m_NXStrips[DetID];
 
+    // TODO: replace magic numbers for cross-talk by reading parameters from file
+    double MeanElectricField = BiasVoltage / Thickness; // unit: V/cm
```
fhagemann: I'm using the bias voltages to determine the mean electric field in the detector, to estimate their charge drift times.

fhagemann (commented):
@zoglauer This PR for me would be ready for review/to merge.

Only one thing. I'm using a lot of physical constants here, but I also need to give a bias voltage in order to estimate the charge drift times. Are these stored somewhere for the DEE to access?

src/MSubModuleChargeTransport.cxx
```diff
@@ -156,9 +158,23 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+  // Define physical constants
+  constexpr double kB = TMath::K(); // unit: J/K
+  constexpr double ElementaryCharge = TMath::Qe(); // unit: C
+  constexpr double IonizationEnergy = 0.00295; // unit: keV
+  constexpr double Epsilon0 = 8.85418781762039e-14; // unit: F/cm
+  constexpr double EpsilonR = 16.0; // in germanium, unitless
+
+  // TODO: Read bias voltage and temperature of the detector from a file
+  constexpr double BiasVoltage = 1050.0; // unit: V
+  constexpr double Temperature = 87.0; // unit: K
```
fhagemann: And the same thing might also apply for the temperature (which might influence how much the charges diffuse)

fhagemann:
Regardless on finding a solution how to read voltages and temperatures:
I don't plan on adding anything new to this PR. Can someone review, so I can work on the feedback today/tomorrow, to get this PR merged? And address how to read voltages/temperatures in a future PR?

src/MSubModuleChargeTransport.cxx
```diff
@@ -167,25 +183,108 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
     // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
     if (ID >= 0 && ID < NXStrips && std::abs(Pos.Y()) <= YWidth/2.0 && std::hypot(Pos.X(), Pos.Y()) <= Radius) {
-      SH.m_ROE.SetStripID(ID);
-      SH.m_IsGuardRing = false;
+
+      // Apply charge sharing based on relative coordinate to the gap of that strip (0 <= X < XPitch)
+      double XFromGap = std::fmod(Pos.X() + XWidth/2.0, XPitch);
+
+      // calculate σ and µ, assuming t = z / v = z / (µ * E)
+      double SigmaX = std::sqrt(2.0 * kB * Temperature * (Pos.Z() + Thickness / 2.0) / (ElementaryCharge * MeanElectricField)); // in cm
```
zoglauer: Add the reference from where the equations are coming in the code

src/MSubModuleChargeTransport.cxx
```diff
@@ -194,27 +293,105 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+    // constexpr double Z2 = 6.710904485197925;
+    // double CrossTalk = A * std::max(1.0 - std::exp(-B1 * (10.0 * Pos.Z() - Z1)), 1.0 - std::exp(B2 * (10.0 * Pos.Z() - Z2)));
+
     // Calculate HV strip ID by rounding down intentionally to avoid truncation towards zero
     // TODO: Confirm the correct strip pitch based on SMEX detector models
     int ID = static_cast<int>(std::floor((Pos.Y() + YWidth/2.0) / YPitch));
 
     // Check for strip ID and if the position is within the allowed strip length or on the guard ring
     // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
     if (ID >= 0 && ID < NYStrips && std::abs(Pos.X()) <= XWidth/2.0 && std::hypot(Pos.X(), Pos.Y()) <= Radius) {
```
zoglauer: Duplicating (or are there any differences in the equations?) the very complicated code for x and y is very error prone - is there any possibility to make that one function?

zoglauer:
I don't see any immediate big issues. Just make sure to track the todo's

src/MSubModuleChargeTransport.cxx
fhagemann: --> 7a7f3d9: Is this what you mean or are you talking about the big code block?

src/MSubModuleChargeTransport.cxx
fhagemann: Done in 7a7f3d9

src/MSubModuleChargeTransport.cxx
fhagemann: Alternatively, I could export everything to a new function, see https://github.com/fhagemann/nuclearizer/commit/d04652f74ae9669eafb563915c1dbdbab2496a23

fhagemann:
Okay, the X/Y code is now exported to a class function to avoid code duplication of the complicated code.

EDIT: I messed up something, will force-push a fix shortly!

fhagemann:
Ok, I fixed the bug I just introduced, and now I am getting the same ROA files I got when running it with the original version before exporting the code to a separate function!
Ready for review :)

src/MSubModuleChargeTransport.cxx
```diff
@@ -241,6 +216,127 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+
+
+void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool isLV) {
+
+  // This function uses strip coordinates (P and Q) instead of X and Y
+  // (P = perpendicular to strip length, Q = along strip length):
+  //   ╔═════════════════════════════════════════════════╗ ↑
+  // P ║                  STRIP CONTACT                  ║ │ Pitch
+  // ↑ ╚═════════════════════════════════════════════════╝ ↓
+  // └→ Q
```
ckierans, line 229: I'm impressed by your ascii art!

src/MSubModuleChargeTransport.cxx
```diff
@@ -241,6 +216,127 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+    m_ChargeTransportHits.push_back(MainSH);
+
+    // create MDEEStripHit for the left NN
+    if (NNLeftStripEnergy > IonizationEnergy) {
+      MDEEStripHit NNLeftSH = SH;
+      NNLeftSH.m_Energy = NNLeftStripEnergy;
+      if (ID > 0) {
+        NNLeftSH.m_ROE.SetStripID(ID - 1);
+        NNLeftSH.m_IsGuardRing = false;
+        // NNLeftSH.m_IsNearestNeighbor = true;
```
ckierans, line 309: I don't think this is where the NN SH should be defined since that should be after the energy for each strip hit is converted to ADC, noised, and then with thresholds taken into account to see what actually triggers, right? So I think this flag will only come in during MSubModuleStripTrigger.

src/MSubModuleChargeTransport.cxx
```diff
@@ -241,6 +216,127 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+        std::erf(b) * (2.0 * std::pow(Eta, 3) + x * (3.0 * std::pow(Eta, 2) - 3.0 * std::pow(Sigma, 2) - std::pow(x, 2))) + 
+        std::erf(a) * (2.0 * std::pow(Eta, 3) - x * (3.0 * std::pow(Eta, 2) - 3.0 * std::pow(Sigma, 2) - std::pow(x, 2))) + 
+        std::exp(- std::pow(b,2)) * std::sqrt(2 / TMath::Pi()) * Sigma * (Eta * x + (2.0 * std::pow(Eta, 2) - 2.0 * std::pow(Sigma, 2) - std::pow(x, 2))) + 
+        std::exp(- std::pow(a,2)) * std::sqrt(2 / TMath::Pi()) * Sigma * (Eta * x - (2.0 * std::pow(Eta, 2) - 2.0 * std::pow(Sigma, 2) - std::pow(x, 2))) 
+      );
+    };
+
+    double MainStripEnergy    = Lambda(Pitch - FromGap) - Lambda(-FromGap);
+    double NNLeftStripEnergy  = Lambda(-FromGap) - Lambda(-Pitch - FromGap);
+    double NNRightStripEnergy = Lambda(2.0*Pitch - FromGap) - Lambda(Pitch - FromGap);
```
ckierans, line 293: Do we ever expect the charge transport to result in the energy deposited across more than 3 strips?

src/MSubModuleChargeTransport.cxx
```diff
@@ -241,6 +216,127 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+
+    // create entry for the main hit
+    MDEEStripHit MainSH = SH;
+    MainSH.m_ROE.SetStripID(ID);
+    MainSH.m_Energy = MainStripEnergy;
+    MainSH.m_IsGuardRing = false;
+    m_ChargeTransportHits.push_back(MainSH);
```
ckierans, line 300: Here you've defined a new DEEStripHit with the above parameters, but you don't retain some of the info that the SH previously had, such as m_SimulatedPosition. This position (or at least the variable m_SimulatedRelativeDepth) will need to be used in the depth SubModule. Maybe I'm misreading how you're clearing and defining events so that this info is still retained for the SH for future submodules to access?

src/MSubModuleChargeTransport.cxx
```diff
@@ -156,65 +158,38 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+  Event->GetDEEStripHitHVListReference().clear();
+  for (MDEEStripHit& SH: m_ChargeTransportHits) {
+    Event->AddDEEStripHitHV(SH);
+  }
+
+  m_ChargeTransportHits.clear();
+
   // Merge hits:
+  // TODO: how to deal with flags like "m_IsNearestNeighbor" etc. ?
   for (auto IterLV1 = LVHits.begin(); IterLV1 != LVHits.end(); ++IterLV1) {
```
ckierans, line 193: Above you've cleared the Event->GetDEEStripHitHVListReference() and then added all of the SH you recovered from the RunChargeTransportForHit function, but now you're iterating through LVHits to merge strip hits, but LVHits was previously defined with the old non-charge-transported list of strip hits. Am I missing something here?

src/MSubModuleChargeTransport.cxx
```diff
@@ -241,6 +216,127 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+  int NStrips        = isLV ? m_NXStrips[DetID] : m_NYStrips[DetID];
+
+  // Express coordinates of the hit in local strip coordinates
+  MVector Pos = SH.m_SimulatedPositionInDetector;
+  double P    = isLV ? Pos.X() : Pos.Y();
+  double Q    = isLV ? Pos.Y() : Pos.X();
+  double ΔZ   = isLV ? Pos.Z() + Thickness / 2.0 : Thickness / 2.0 - Pos.Z();
+
+  // Calculate strip ID by rounding down intentionally to avoid truncation towards zero
+  int ID = static_cast<int>(std::floor((P + PWidth/2.0) / Pitch));
```
ckierans, line 250: You'll eventually need to include the mask metrology information when calculating the Strip ID from the position. Mark this as a TODO in the code for now.

src/MSubModuleChargeTransport.cxx
```diff
@@ -241,6 +216,127 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
...
+
+    // create entry for the main hit
+    MDEEStripHit MainSH = SH;
+    MainSH.m_ROE.SetStripID(ID);
+    MainSH.m_Energy = MainStripEnergy;
+    MainSH.m_IsGuardRing = false;
+    m_ChargeTransportHits.push_back(MainSH);
+
+    // create MDEEStripHit for the left NN
+    if (NNLeftStripEnergy > IonizationEnergy) {
```
ckierans, line 303: Is this logic right? Why does the amount of charge deposited on the neighbor strip depend on the ionization energy?

ckierans (commented):
The code looks good, Felix! I have a few questions about the logic, and mostly they're for me to understand what's actually happening. After you confirm none of my comments are an issue, I'm okay with merging this knowing it's still a work in progress.

src/MSubModuleChargeTransport.cxx
fhagemann: My thought behind this: we create one electron-hole pair per `IonizationEnergy`, and if the energy collected by a strip is less than that, that would mean it would have collected less than one electron/hole. I can also change it to something else.

src/MSubModuleChargeTransport.cxx
fhagemann: I agree. Should I move this to `MSubModuleStripTrigger` already within this PR?

src/MSubModuleChargeTransport.cxx
fhagemann: I see, and I can check this.. My understanding was that by defining `MainSH` via `SH`, all the previous information from the simulation would still remain saved, and I would just be "adding" the new information from the charge transport.

fhagemann: Another question regarding your comment:
How should we deal with events with multiple hits within one strip, that would get merged later on? Should they keep the simulated position of the dominant hit? Or of the hit closest to the contact?

src/MSubModuleChargeTransport.cxx
fhagemann: I see what you mean...
`LVHits` is "just" a reference to the list (rather than the list itself), so even if it's cleared and re-populated after being defined, iterating through `LVHIts` should still work (or at least I'm getting reasonable spectra when doing so).
But I might also be wrong.

src/MSubModuleChargeTransport.cxx
fhagemann: I will add the TODO.

src/MSubModuleChargeTransport.cxx
fhagemann: I'm not getting significant contribution to next-to-nearest neighbors for the energies I've look at so far (~1MeV in a single hit), but I can expand this also to next-to-nearest neighbors.

Also: if we run simulations with `Simple` detectors rather than clustering hits with `Voxel3D`, we might not expect individual hits with energies way higher than that, but rather a bunch of lower-energetic sub-hits.

src/MSubModuleChargeTransport.cxx
ckierans: You can wait till you start working through the MSubModuleStripTrigger code. You can add a TODO in that file for now if you want.

fhagemann:
One open question I had looking at this was how to define „event time“ (with setCL etc.), see also #69

src/MSubModuleChargeTransport.cxx
ckierans: Okay, I guess I was thinking that since the ionization occurs before the charge transportation then you could get a fraction of the ionization energy on a neighboring strip. I'm fine with your logic, I mostly am just making sure I understood what you are doing..

src/MSubModuleChargeTransport.cxx
ckierans: You're probably right. References always get me.

src/MSubModuleChargeTransport.cxx
ckierans: So what I'm hearing is that if you keep it just as nearest neighbors then there might be a small loss of energy, but that'd probably be below a reasonable detectable energy so doesn't matter. Maybe just add a comment in the code to show you've thought this through?

src/MSubModuleChargeTransport.cxx
ckierans: You're right about defining MainSH via sH. I missed that somehow.

Good question about the position from events with multiple hits! Are you referring to two interactions that share the same LV strip, but different HV strips, for example? If I'm understanding your code right, you're combining the energy of multiple hits per strip in your iteration starting at line 193. So you no longer have two separate hits clearly defined with separate positions? If so, I'm wondering if this is the right place to make this combination or if we can keep the hits separate until we get to the DepthReadout where we can handle the timing of each hit properly. 
We'd want this to mimic the hardware as much as possible. I assume the strip hit is assigned the timing of the closest hit as long as it's above the fast threshold, right? So I don't think you could easily determine the correct position to use here without doing the inverse depth calibration to get timing for each strip hit and have fast threshold knowledge.

ckierans (commented):
My only open question is about combining strip hits for instances of multiple hits per strip. Combining them at this stages means we lose position information for these hits. Maybe this is a fine temporary solution since the number of events that have multiple hits per strip is low (but definitely non-zero!), but as we continue to work through the rest of the submodules, we'll need to make sure we find a way to handle these events in the most realistic way.

In conclusion, if you want to move on, I'm fine merging this for now, but let's keep thinking about these multiple-hit-per-strip events.

src/MSubModuleChargeTransport.cxx
fhagemann: --> 27ebf2a

src/MSubModuleChargeTransport.cxx
fhagemann: I thought about this again this morning: we might still want to expand this to next-to-nearest neighbors, just because of the fact that the nearest neighbors might surpass the energy threshold to actually be considered triggered events, and then the next-to-nearest neighbor might still be read out as "nearest-neighbor" of the triggered strip.

src/MSubModuleChargeTransport.cxx
fhagemann: This was code that @zoglauer introduced in his initial DEE implementation (c527242d4e38adafb413b03146639b17ef085b66) and that I didn't touch so far.
From what you are saying, it might make sense to combine those hits only once the simulated information (energy, position) of the individual hits are not needed anymore, maybe in `MSubModuleStripReadout` or `MSubModuleDEEOutput` (?).

The fast shaper is supposed to trigger on the first peak above the fast threshold, which should correspond to the earliest timing of electrons and holes arriving. This situation gets a bit more complicated in charge-sharing signals, where the fast shaper signal is not necessarily "just a peak", but might exhibit somewhat of a peak-dip structure, resulting in slightly earlier triggering than what is expected from the electron- and hole-arrival times. This might become particularly relevant, if we want to be able to simulate things like NN timing for bump corrections.

src/MSubModuleChargeTransport.cxx
fhagemann: A fraction of the ionization energy (we're talking about 2.95eV = 0.00295keV here) would correspond to a fraction of an electron/hole, which would make no physical sense. Assuming that the noise level on the slow-shaper might be of the order of 1mV = 1keV in the slow shaper and ~5mV = 5keV in the fast shaper, I think choosing 0keV or ionization energy here should really make no difference.

I can revert this to 0keV, and/or also add a TODO.

ckierans: Agreed, it's a small difference. Leave this as is. No need to add a TODO.

src/MSubModuleChargeTransport.cxx
ckierans: Yes, I think the hits should be combined either in the DepthReadout or in the DEEOutput. The StripReadout only handles the inverse energy calibration, but we'll need the hit positions separate for the inverse depth calibration for an accurate handling of the timing for each hit. Again, I'm fine with this as is for now, but keep this in mind as we progress with the fidelity of the DEE.

ckierans:
I'm ready to merge this when you are.

fhagemann:
This can also be merged from my side.
I can proceed with opening new issues for the TODOs introduced in this PR.

PR #86, opened 2026-01-13
[DEE] Support `MDVoxel3D` detectors in `MSubModuleChargeTransport`

fhagemann:
@zoglauer I'm opening this PR just to get some quick feedback if this is what you envisioned. 
This PR will need to go hand in hand with some mass model updates to also migrate from `MDStrip3D` to `MDVoxel3D`.

### DIFF
--- a/include/MSubModuleChargeTransport.h
+++ b/include/MSubModuleChargeTransport.h
@@ -24,6 +24,7 @@
 #include "MGlobal.h"
 #include "MSubModule.h"
 #include "MDStrip3D.h"
+#include "MDVoxel3D.h"
 
 // Forward declarations:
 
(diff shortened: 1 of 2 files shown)


zoglauer:
OK.
We have to separate two models: We simulate with a high resolution voxel detector, do charge transport, and then "rebin" it into our normal 64-strip detector again.
What info do you need in this class? The high-resolution simulated voxel information, or the real strip information? My guess is the latter?

fhagemann:
In this module, I need the detector crystal and strip dimensions (width and thickness of the crystal, strip pitch...), to attribute the hits to the correct strips (or guard ring), knowing how close a hit was to the strip gap and how much charge was shared. Is this what you mean by "real strip information"?

zoglauer:
Yes. real strips not voxels.

fhagemann:
Yes, I need the real strip information here then.

zoglauer:
Then what you had previously is OK.
The key thing is to modify the geometry to contain the voxel detector, simulated voxels, read in the voxel data, and use that for the charge transport sim

fhagemann:
Closing this, as we will be using Voxel3D detectors in cosima, but then still using the Strip3D definition in the geometry for the DEE (for now).

PR #90, opened 2026-01-21
Adding Strip Pairing Quality Flags

julianmgerber:
Implementing quality flags for the strip pairing module instead of just filtering out events.

Thresholds/filters that are now a quality flag:
1. If there are more than 6 strip hits on one side
2. If the best LV or HV combination has more than 5 strip groupings
3. If the best strip pairing chi square is > 25
4. If the LV/HV energies are further apart than 2.5 sigma
5. Multiple hits on a single strip

Points 1-4 are all included in the new StripPairingQualityFlag variable and each have their own warning message attached to them. The flags for multiple hits on a single strip however are defined on the MHit level (as opposed to the event level) and are not included in the StripPairingQualityFlag category.

### DIFF
--- a/include/MModuleLoaderMeasurementsFITS.h
+++ b/include/MModuleLoaderMeasurementsFITS.h
@@ -100,7 +100,7 @@ class MModuleLoaderMeasurementsFITS : public MModuleLoaderMeasurements
   long m_TotalRows;
 
   //! Batch size for reading FITS data
-  static const long m_BatchSize = 100;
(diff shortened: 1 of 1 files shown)


zoglauer:
That one conflicts with @ckierans changes - let Carolyn do her changes first and then update. Then i will review

ckierans:
@julianmgerber the other PR handling QA and BD flags was merged, so you should be able to make progress on this one now.

PR #91, opened 2026-01-27
ADD: L0 binary saver module for DD packets

morpinma-wing:
Hi Andreas,                                                                                                               
I have drafted the sim-to-L0 saver. This is not be perfect, and I'm hoping you can give me feedback such as am I on the right track? Is this completely wrong or off-track?
                                                                                                                            
  I also have a few questions:                                                                                              
  1. What does GetCL() return?                                                                                              
  2. Where can I get the event type?                                                                                        
                                                                                                                            
I appreciate any feedback you have. Thank you!


zoglauer:
You have to make the truncated time yourself from the time and you have to make the event type yourself, GetCL() was from balloon times.

zoglauer:
I think you will have to make the time code from GetTime()

zoglauer:
Code looks good. We just need to test it....

mwy-wing:
How do I make event type myself?... I don't understand this part. Can you clarify more on this?

mwy-wing:
I made some new changes, including change from GetCL() to GetTime(), and also add a function that detect the event type from event.

mwy-wing:
I tested with the file Felix gave me, and it looks good to me. But for some reason, I see that GetTime returns the same time as GetCL. I think it's because it's a simulation file? (maybe?)

fhagemann:
> I tested with the file Felix gave me, and it looks good to me. But for some reason, I see that GetTime returns the same time as GetCL. I think it's because it's a simulation file? (maybe?)

Don't worry too much about the values in `GetCL` and `GetTime` from simulation output for now too much..
We are still finalizing the convention what time values to store, so these values are all preliminary right now.

PR #99, opened 2026-02-23
Strip Pairing Flags

julianmgerber:
Summary of changes:
- Added strip pairing quality flags for the following things:
  - If the best combination has more than 5 strip groupings (ie. potential hits) on one side
  - If the best strip pairing chi square is > 25
  - If the LV/HV energies are further apart than 2.5 sigma
  - If there are multiple hits on a single strip
  - If the best strip pairing leaves any strips unpaired
    - Read out the total unpaired energy in QA flag
- Split the MHit charge sharing flag into two flags (one for LV and one for HV)
  - Assigned hits charge sharing flag in strip pairing
- Added a guard ring hit flag to MHit
  - At end of strip pairing, check if any paired hit contains a guard ring strip and, if so, assign the guard ring hit flag
  - Read out GR hit info in QA flag (detector ID + energy)
- Implemented options GUI so user can select maximum number of strip hits on one side that can be paired

### DIFF
--- a/include/MHit.h
+++ b/include/MHit.h
@@ -108,10 +108,18 @@ class MHit
 	//! get m_StripHitMultipleTimesY
 	bool GetStripHitMultipleTimesY() const { return m_StripHitMultipleTimesY; }
 
-	//! set charge sharing flag
-	void SetChargeSharing(bool chargeSharing) {m_ChargeSharing = chargeSharing; }
-	//! get m_ChargeSharing
-	bool GetChargeSharing() const { return m_ChargeSharing; }
+	//! set charge sharing flag for LV side
+	void SetChargeSharingLV(bool chargeSharingLV) {m_ChargeSharingLV = chargeSharingLV; }
+	//! get m_ChargeSharingLV
+	bool GetChargeSharingLV() const { return m_ChargeSharingLV; }
+    //! set charge sharing flag for HV side
+    void SetChargeSharingHV(bool chargeSharingHV) {m_ChargeSharingHV = chargeSharingHV; }
+    //! get m_ChargeSharingHV
+    bool GetChargeSharingHV() const { return m_ChargeSharingHV; }
+    //! Keeping general charge sharing flags because Greedy strip pairing relies on it
+    void SetChargeSharing(bool chargeSharing) {m_ChargeSharing = chargeSharing; }
+    //! get m_ChargeSharing
+    bool GetChargeSharing() const { return m_ChargeSharing; }
 	//! set m_NoDepth
 	void SetNoDepth(bool X = true) { m_NoDepth = X;}
 	//! get m_NoDepth
@@ -184,7 +192,9 @@ class MHit
 	bool m_StripHitMultipleTimesY = false;
 
 	//! true if hit contains charge sharing
-	bool m_ChargeSharing;
+    bool m_ChargeSharing;
+	bool m_ChargeSharingLV;
+    bool m_ChargeSharingHV;
 
 	//! true if depth is invalid, either because the pixel depth was uncalibrated, the hit was mapped too far out of the detector,or there was no timing data
 	bool m_NoDepth;
--- a/include/MHit.h
+++ b/include/MHit.h
@@ -93,6 +93,11 @@ class MHit
 	void SetCrossTalkFlag(bool PossibleCrossTalk) {m_PossibleCrossTalk = PossibleCrossTalk;}
 	//! get cross talk flag value
 	bool GetCrossTalkFlag() const { return m_PossibleCrossTalk; }
+    
+    //! Set guard ring hit flag
+    void SetGuardRingHitFlag(bool GuardRingHit) {m_GuardRingHit = GuardRingHit;}
+    //! Get guard ring hit flag
+    bool GetGuardRingHitFlag() const { return m_GuardRingHit; }
 
 	//! set charge loss flag
 	void SetChargeLossFlag(bool PossibleChargeLoss) {m_PossibleChargeLoss = PossibleChargeLoss;}
@@ -185,6 +190,8 @@ class MHit
 	bool m_PossibleCrossTalk;
 	//! Flag: possible charge loss
 	bool m_PossibleChargeLoss;
+    //! Flag: hit containing guard ring strip
+    bool m_GuardRingHit;
 
 	//! true if hit contains strip that was hit multiple times on X
 	bool m_StripHitMultipleTimesX = false;
--- a/src/MModuleStripPairingMultiRoundChiSquare.cxx
+++ b/src/MModuleStripPairingMultiRoundChiSquare.cxx
@@ -352,6 +352,10 @@ tuple<vector<vector<unsigned int>>, vector<vector<unsigned int>>, double> MModul
     for (unsigned int hv = 0; hv < Combinations[d][1].size(); ++hv) {
       
       unsigned int MinSize = min(Combinations[d][0][lv].size(), Combinations[d][1][hv].size());
+        
+      if (abs(long(Combinations[d][0][lv].size()) - long(Combinations[d][1][hv].size())) > 1) {
+        continue;
+      }
 
       bool MorePermutations = true;
       while (MorePermutations == true) {
--- a/src/MModuleStripPairingMultiRoundChiSquare.cxx
+++ b/src/MModuleStripPairingMultiRoundChiSquare.cxx
@@ -567,6 +567,7 @@ bool MModuleStripPairingMultiRoundChiSquare::CreateHits(unsigned int d, MReadOut
     if (AllAdjacentHV == false && AllAdjacentLV == true) {
       //cout<<"Multiple hits on single LV strip"<<endl;
       bool MultipleHitsOnLV = true;
+      Event->SetStripPairing_QualityFlag("Event contains multiple hits on single LV strip");
 
       // Assign hit energy based on energy measured on HV side
       for (unsigned int sh = 0; sh < BestHVSideCombo[h].size(); ++sh) {
@@ -588,7 +589,6 @@ bool MModuleStripPairingMultiRoundChiSquare::CreateHits(unsigned int d, MReadOut
         Hit->SetLVEnergy(LVEnergy);
         Hit->SetHVEnergy(HVEnergy);
         Hit->SetStripHitMultipleTimesX(MultipleHitsOnLV);
-        Event->SetStripPairing_QualityFlag("Event contains multiple hits on single LV strip");
         Event->AddHit(Hit);
           
 
@@ -614,6 +614,7 @@ bool MModuleStripPairingMultiRoundChiSquare::CreateHits(unsigned int d, MReadOut
     else if (AllAdjacentLV == false && AllAdjacentHV == true) {
       // cout<<"Multiple hits on single HV strip"<<endl;
       bool MultipleHitsOnHV = true;
+      Event->SetStripPairing_QualityFlag("Event contains multiple hits on single HV strip");
 
       // Assign hit energy based on energy measured on LV side
       for (unsigned int sh = 0; sh < BestLVSideCombo[h].size(); ++sh) {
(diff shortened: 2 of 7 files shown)


include/MHit.h
```diff
@@ -108,10 +113,18 @@ class MHit
...
-	bool GetChargeSharing() const { return m_ChargeSharing; }
+	//! set charge sharing flag for LV side
+	void SetChargeSharingLV(bool chargeSharingLV) {m_ChargeSharingLV = chargeSharingLV; }
+	//! get m_ChargeSharingLV
+	bool GetChargeSharingLV() const { return m_ChargeSharingLV; }
+    //! set charge sharing flag for HV side
+    void SetChargeSharingHV(bool chargeSharingHV) {m_ChargeSharingHV = chargeSharingHV; }
+    //! get m_ChargeSharingHV
+    bool GetChargeSharingHV() const { return m_ChargeSharingHV; }
+    //! Keeping general charge sharing flags because Greedy strip pairing relies on it
```
ckierans, line 124: Andreas was proposing we remove the Greedy algorithm code since we're not using it. Any objections? I can open a separate PR and remove these flags at that point.

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -272,9 +272,7 @@ bool MModuleStripPairingMultiRoundChiSquare::EventSelection(MReadOutAssembly* Ev
   for (unsigned int d = 0; d < StripHits.size(); ++d) { // Detector loop
     for (unsigned int side = 0; side <= 1; ++side) { // Side loop
       if (StripHits[d][side].size() > MaxStripHits) {
-        Event->SetStripPairingError("More than 6 hit strips on one side");
-        Event->SetAnalysisProgress(MAssembly::c_StripPairing);
-        return false;
+        Event->SetStripPairing_QualityFlag("More than 6 hit strips on one side");
```
ckierans: I would think this circumstance would still lead to a BD flag, not a QA flag? Are you able to pair strips when you have more than 6 strip hits on one side?

Is this looser requirement related to you removing the MaxCombination restriction in the FindAllCombinations function below?

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -352,18 +350,9 @@ tuple<vector<vector<unsigned int>>, vector<vector<unsigned int>>, double> MModul
...
-      }
-
+      
       unsigned int MinSize = min(Combinations[d][0][lv].size(), Combinations[d][1][hv].size());
 
-      // Skip pairing if either side has more than 5 sets of strips
-      if (max(Combinations[d][0][lv].size(), Combinations[d][1][hv].size()) > MaxCombinations) {
-        continue;
-      }
-
```
ckierans, line 366: I'm curious as to why you no longer need the selection on lv/hv strip combos differing and the max combo size. I don't see you accounting for this elsewhere.

src/MModuleStripPairingMultiRoundChiSquare.cxx
ckierans: if (AllAdjacent == true) {

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -604,13 +603,18 @@ bool MModuleStripPairingMultiRoundChiSquare::CreateHits(unsigned int d, MReadOut
...
+          Hit->SetChargeSharingLV(true);
+        }
       }
     }
 
     // Multiple hits on HV side
     else if (AllAdjacentLV == false && AllAdjacentHV == true) {
       // cout<<"Multiple hits on single HV strip"<<endl;
       bool MultipleHitsOnHV = true;
+      Event->SetStripPairing_QualityFlag("Event contains multiple hits on single HV strip");
```
ckierans: Do we need a different QA flag for multiple hits on LV vs HV strip? What do we gain from this distinction?

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -660,9 +668,7 @@ bool MModuleStripPairingMultiRoundChiSquare::CreateHits(unsigned int d, MReadOut
 
   // One last quality selection based on total event energies
   if ((EnergyTotal > max(LVEnergyTotal, HVEnergyTotal) + 2.5 * max(LVEnergyResTotal, HVEnergyResTotal) || EnergyTotal < min(LVEnergyTotal, HVEnergyTotal) - 2.5 * max(LVEnergyResTotal, HVEnergyResTotal))) {
-    Event->SetStripPairingError("Strips not pairable wihin 2.5 sigma of measured energy");
-    Event->SetAnalysisProgress(MAssembly::c_StripPairing);
-    return false;
+    Event->SetStripPairing_QualityFlag("Strips not pairable wihin 2.5 sigma of measured energy");
```
ckierans, line 677: Is there a sigma cutoff that we truly would want to filter out?

From my previous checks with 662 keV data, I found this is only 0.04% of events, so shouldn't be a major concern, just wondering where were should actually draw the line.

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -765,21 +771,38 @@ bool MModuleStripPairingMultiRoundChiSquare::AnalyzeEvent(MReadOutAssembly* Even
     }
     // Flag events with a reduced chi square > 25
     else if (BestChiSquare > 25) {
-      Event->SetStripPairingError("Best reduced chi square is not below 25");
-      Event->SetAnalysisProgress(MAssembly::c_StripPairing);
-      return false;
+      Event->SetStripPairing_QualityFlag("Best reduced chi square is not below 25");
```
ckierans: Can you report the chi square in the QA message?

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -765,21 +771,38 @@ bool MModuleStripPairingMultiRoundChiSquare::AnalyzeEvent(MReadOutAssembly* Even
...
     }
 
     // Assign the best reduced chi square to the event
     Event->SetStripPairingReducedChiSquare(BestChiSquare);
+    
+    // Check if size of LV or HV combination exceeds maximum
+    if (max(BestLVSideCombo.size(), BestHVSideCombo.size()) > MaxCombinations) {
+      Event->SetStripPairing_QualityFlag("Best strip pairing contains more than 5 strip groupings (hits)");
```
ckierans: Now i see how your removed the checks above. Good idea to run through the pairing and just see if it works! What fraction of events have more than 5 hits?

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -765,21 +771,38 @@ bool MModuleStripPairingMultiRoundChiSquare::AnalyzeEvent(MReadOutAssembly* Even
...
+    }
+    
+    // Flag event if more than one set of strips is left unpaired
+    if (abs(long(BestLVSideCombo.size()) - long(BestHVSideCombo.size())) > 1) {
+      Event->SetStripPairing_QualityFlag("Best strip pairing leaves more than one grouping of strips unpaired");
```
ckierans: Hmm, this seems concerning. Does this mean there are triggered strips that don't end up in a hit? I think this means it's a BD event? 

What fraction of events get this flag?

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -765,21 +771,38 @@ bool MModuleStripPairingMultiRoundChiSquare::AnalyzeEvent(MReadOutAssembly* Even
...
     if (PopulateHits == false) {
       return false;
     }
-
+      
+    // Check for any hits containing guard ring strips
+    for (unsigned int h = 0; h < Event->GetNHits(); h++) {
+      for (unsigned int sh = 0; sh < Event->GetHit(h)->GetNStripHits(); sh++) {
+        if (Event->GetHit(h)->GetStripHit(sh)->GetStripID() == 64) {
+          Event->GetHit(h)->SetGuardRingHitFlag(true);
```
ckierans, line 824: Good! Thanks for adding this!

ckierans (commented):
Looks good, Julian! Just a few comments below.

ckierans:
Now that you're allowing these "lesser quality" events through, how does the ARM change in mimrec? How does the number of events that are reconstructed change?

zoglauer:
"Turning >6 strip hits back into a BD flag due to runtime"

Sorry, but you cannot do this by default - this would suppress high energy events!
Could you instead add an entry into the IU, to choose up to how many strips per detector the strip pairing should run?

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: I think at one point I was counting them separately, but it's probably not necessary. I just made it into a single flag

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: Done!

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: I set up a GUI for choosing the maximum number of strip hits allowed on one side. I think it should still be able to pair with a larger maximum, but I have noticed that it really slows down above 6 strip hits

julianmgerber:
> Now that you're allowing these "lesser quality" events through, how does the ARM change in mimrec? How does the number of events that are reconstructed change?

@ckierans With all the quality flags, the new ARM is 9.48 +/- 0.12 deg compared to the original 9.19 +/- 0.13 deg. Within 640-675 keV, there are 294304 reconstructed events (1.85% of total events) compared to the original 253478 (1.60%)

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: For streaming these GR hits to an evta file, are you thinking that the specific hit itself would get a flag as opposed to the general "BD GR Veto" flag?

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: Only 1250 events get this flag (out of >15 million events).

But actually since the condition is max(BestLVSideCombo, ...), then what can happen is one combination has >5 but the other only has 4, so there are only 4 hits and one strip gets left unpaired.

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: 27,003 out of 15.8 million events have this flag.
Should there be a BD flag for even just one strip (or strip grouping) left unpaired? I'm thinking this type of quality flag might be able to account for any noisy strips that weren't filtered out by the slow threshold.

ckierans: Yes, if you could add a quality flag and record the energy that was not paired, that'd be great.

src/MModuleStripPairingMultiRoundChiSquare.cxx
ckierans: Each Event will should still get the "BD GR Veto" flag (from the m_GuardRingVeto flag), since we usually want to filter these out on the event basis. But within these BD events, we'd like the record the GR Hit information somehow. We don't have position info for GR Hits, but we do have energy and Detector ID. 

It's not obvious to me the best way to save the GR Hit info in the Evta file. Here are my current ideas:
- these parameters can be appended to the BD GR Veto message
- we could define an additional line "HT GR" instead of "HT 3" with the energy information and Detector ID
- we could just track the GR Hit information in a QA flag: "QA GR Hit: Det ID # and energy XX"

In the final L1b/L2 FITS format we only record the energy and position of a good hit and not the Detector ID, so I think we'll ultimately keep this GR Hit information in the BD(/QA) Flag message for each event and that makes me lean towards the 3rd option above.

julianmgerber: Ok, I went with the third option!

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: Done!

ckierans:
Thanks for comparing the ARMs and the number of events with this new QA flag. Do you think the majority of this ARM degradation comes from the chi-square cut? It's obvious we'll need filtering implemented in mimrec sooner than later. We'll need to think about the best way to do this, but I could imagine having another option in the "Quality" tab under "Event selection" to filter on the chi-square of the strip pairing, if this is the most impactful parameter. Let's discuss at the next CDEE meeting. 

You should think about if there are other QA parameters we should be filtering on, and we should hear from Andreas if he wants this Strip Pairing Quality Factor to be something that is reported on a scale from 0-100 like the other quality factors in mimrec.

PR #100, opened 2026-02-26
ADD: Support for loading version 2 of the HDF hit info

### LINKED ISSUE #57: MModuleLoaderMeasurementsHDF Number of Events is 0
When running Nuclearizer the MModuleLoaderMeasurementsHDF prints out number of events on file at the end, but all I'm getting is 0 when that's obviously not the case:

```

MModuleLoaderMeasurementsHDF: 
  * all events on file: 0
  * good events on file: 0
###################
AWL depth cal stats
###################
Good hits: 1559674
Number of hits missing calibration coefficients: 16624
Number of hits too far outside of detector: 176847
Number of hits missing timing information: 0
Number of hits with strips hit multiple times: 0
Number of hits with non-adjacent strip hits: 0
Number of hits with too many strip hits: 0
Number of hits with no strip hits on one or both sides: 107
Number of hits with null strip hits: 0
Number of hits 0 energy on a strip hit: 0

Nuclearizer: Analysis finished in 22.6882s

Summary: 
Spent 4.3 sec analyzing (vs. 17.7 sec sleeping) in module "Measurement loader for HDF files" utilizing 1 instance and processed 2332091 events.
Spent 9.7 sec analyzing (vs. 142.2 sec sleeping) in module "NearestNeighbor" utilizing 7 instances and processed 2332090 events.
Spent 1.4 sec analyzing (vs. 20.7 sec sleeping) in module "TAC Calibration" utilizing 1 instance and processed 2332090 events.
Spent 2.2 sec analyzing (vs. 20.0 sec sleeping) in module "Universal energy calibrator" utilizing 1 instance and processed 2332090 events.
Spent 10.1 sec analyzing (vs. 12.0 sec sleeping) in module "Strip pairing - chi square approach" utilizing 1 instance and processed 2332090 events.
Spent 7.1 sec analyzing (vs. 15.1 sec sleeping) in module "Depth calibration 2024" utilizing 1 instance and processed 2332090 events.
Spent 6.5 sec analyzing (vs. 15.7 sec sleeping) in module "Save events (roa, dat, or evta format)" utilizing 1 instance and processed 2332090 events.
```

fhagemann:
This is my attempt to implement a HDF loader for the newest version 2 of the HDF hit into.
I'm not claiming that this is perfect, but it seems to give something reasonable at least..
It also fixes #57.

I developed this based on a file that was taken with `HDF v2.3` and `GSE v7.2.2` on the FM assembly at SSL.

The changes between versions 1 and 2 addressed in this PR are:
- **where the HDFVersion is saved in the file**
  | Version 1 | Version 2 |
  |----------|----------|
  | `HDFVersion` is its own dataset within the file. | `HDFVersion` is an attribute of the `Events` dataset. |
  |  <pre>🗂️ HDF5.File (version 1)<br>  ├─ 🏷️ CLASS<br>  ├─ 🏷️ FILTERS<br>  ├─ 🏷️ PYTABLES_FORMAT_VERSION<br>  ├─ 🏷️ TITLE<br>  ├─ 🏷️ VERSION<br>  ├─ 🔢 Buffers<br>  ├─ 🔢 Config<br>  ├─ 🔢 GSEVersion<br>  ├─ 🔢 HDFVersion<br>  │  ├─ 🏷️ CLASS<br>  │  ├─ 🏷️ FIELD_0_FILL<br>  │  ├─ 🏷️ FIELD_0_NAME<br>  │  ├─ 🏷️ NROWS<br>  │  ├─ 🏷️ TITLE<br>  │  └─ 🏷️ VERSION<br>  ├─ 🔢 Headers<br>  └─ 🔢 Hits</pre> | <pre>  🗂️ HDF5.File (version 2)<br>  ├─ 🔢 ACSHits<br>  ├─ 🔢 Buffers<br>  ├─ 🔢 DIBCoincidence<br>  ├─ 🔢 DetectorHits<br>  ├─ 🔢 DetectorLiveTime<br>  ├─ 🔢 EventIndices<br>  ├─ 🔢 Events<br>  │  ├─ 🏷️ BackendVersion<br>  │  ├─ 🏷️ Config<br>  │  ├─ 🏷️ Firmware<br>  │  ├─ 🏷️ GSEVersion<br>  │  └─ 🏷️ HDFVersion<br>  ├─ 🔢 FEEHits<br>  ├─ 🔢 H&S<br>  ├─ 🔢 Headers<br>  ├─ 🔢 RTBHousekeeping<br>  ├─ 🔢 SCBHousekeeping<br>  └─ 🔢 SinglesCounts</pre> |

  --> If no dataset `HDFVersion` exists, check for the existence of `Events` and attribute `HDFVersion` therein.
 
- **Hit and event information are now stored in different datasets**
  | Version 1 | Version 2 |
  |----------|----------|
  | Both, hit and event information was stored in the dataset `Hits`. | The hit information is saved in the dataset `FEEHits` and the event information is saved in the dataset `Events`. Hit entries in `FEEHits` have a field `event_index` which denotes which "row" in the `Events` dataset correspond to the given hit. |

  For clarification:
  - Hit information = strip ID, energy data, timing data, hit type, timing type
  - Event information = event ID, timecode, GSE timecode, SPW timecode, number of hits, ...
  
  --> use the same buffer logic for **hit information**, but read in the **event information** as a whole.

Right now, the handling of `SPWTimeCode` (spacewire time code) is commented out to avoid undeclared variable warnings.

I would be very grateful for feedback! :)

### DIFF
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -257,6 +257,9 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
   //! The file name of the strip map
   MString m_FileNameStripMap;
 
+  //! The map which ASICs are HV/LV for HDFv2
+  vector<map<bool, vector<bool>>> m_ASICPolarities;
+
   //! The strip map
   MStripMap m_StripMap;
   
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -112,7 +112,8 @@ struct MHDFStripHit_V2 {
 struct MHDFEvent_V2_0 {
   uint16_t m_EventID;
   uint64_t m_TimeCode;
-  double   m_GSETimeCode;
+  // Nanoseconds since epoch
+  uint64_t m_GSETimeCode;
   uint8_t  m_Hits;
   uint16_t m_Bytes;
   uint8_t  m_EventType;
@@ -123,8 +124,10 @@ struct MHDFEvent_V2_0 {
 struct MHDFEvent_V2_2 {
   uint16_t m_EventID;
   uint64_t m_TimeCode;
-  double   m_GSETimeCode;
-  double   m_SPWTimeCode;
+  // Nanoseconds since epoch
+  uint64_t m_GSETimeCode;
+  // Nanoseconds since start of acquisition
+  uint64_t m_SPWTimeCode;
   uint8_t  m_Hits;
   uint16_t m_Bytes;
   uint8_t  m_EventType;
@@ -278,7 +281,6 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
 
   //! Current hit number in the file
   hsize_t m_CurrentHit;
-  unsigned int m_CurrentEvent;
 
   //! Number of event ID roll-overs:
   unsigned int m_NumberOfEventIDRollOvers;
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -1,7 +1,7 @@
 /*
  * MModuleLoaderMeasurementsHDF.h
  *
- * Copyright (C) by Andreas Zoglauer.
+ * Copyright (C) by Andreas Zoglauer, Felix Hagemann.
  * All rights reserved.
  *
  * Please see the source-file for the copyright-notice.
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -39,13 +39,14 @@ using namespace H5;
 
 
 //! The versions of the strip hits stored in the HDF file
-enum class MHDFStripHitVersion { V1_0, V1_2 };
+enum class MHDFStripHitVersion { V1_0, V1_2, V2_2 };
 
 //! And a streamer for it
 inline ostream& operator<<(ostream& os, MHDFStripHitVersion Version) {
   switch (Version) {
     case MHDFStripHitVersion::V1_0: return os<<"1.0";
     case MHDFStripHitVersion::V1_2: return os<<"1.2";
+    case MHDFStripHitVersion::V2_2: return os<<"2.2";
     default: return os<<"Unknown";
   }
 }
@@ -90,6 +91,28 @@ struct MHDFStripHit_V1_2 {
   uint16_t m_EnergyDataHighGain;
   uint16_t m_TimingData;
   uint8_t  m_Pad;
+  uint8_t  m_Hits;
+  uint16_t m_Bytes;
+  uint8_t  m_EventType;
+  uint8_t  m_CRC;
+};
+
+//! Version 2.2 of the HDF5 hit info
+struct MHDFStripHit_V2_2 {
+  uint32_t m_EventIndex;
+  uint8_t  m_HitType;
+  uint8_t  m_TimingType;
+  uint16_t m_StripID;
+  uint16_t m_EnergyData;
+  uint16_t m_TimingData;
+};
+
+//! Version 2.2 of the HDF5 event info
+struct MHDFEvent_V2_2 {
+  uint16_t m_EventID;
+  uint64_t m_TimeCode;
+  double   m_GSETimeCode;
+  double   m_SPWTimeCode;
   uint8_t  m_Hits;
   uint16_t m_Bytes;
   uint8_t  m_EventType;
@@ -191,9 +214,11 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
   
   //! The HDF5 data set
   DataSet m_HDFDataSet;
+  DataSet m_EventDataSet;
 
   //! The HDF5 compond data type
   CompType m_HDFCompoundDataType;
+  CompType m_EventCompoundDataType;
 
   //! The version of the strip hit structure
   MHDFStripHitVersion m_HDFStripHitVersion;
@@ -212,12 +237,16 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
   vector<MHDFStripHit_V1_0> m_Buffer_1_0;
   //! The MHDFStripHit_V1_2 batch:
   vector<MHDFStripHit_V1_2> m_Buffer_1_2;
+  //! The MHDFStripHit_V2_2 batch:
+  vector<MHDFStripHit_V2_2> m_Buffer_2_2;
+  vector<MHDFEvent_V2_2> m_EventData_2_2;
 
   //! Total number of hits in a file
   hsize_t m_TotalHits;
 
   //! Current hit number in the file
   hsize_t m_CurrentHit;
+  unsigned int m_CurrentEvent;
 
   //! Number of event ID roll-overs:
   unsigned int m_NumberOfEventIDRollOvers;
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -39,13 +39,14 @@ using namespace H5;
 
 
 //! The versions of the strip hits stored in the HDF file
-enum class MHDFStripHitVersion { V1_0, V1_2, V2_2 };
+enum class MHDFStripHitVersion { V1_0, V1_2, V2_0, V2_2 };
 
 //! And a streamer for it
 inline ostream& operator<<(ostream& os, MHDFStripHitVersion Version) {
   switch (Version) {
     case MHDFStripHitVersion::V1_0: return os<<"1.0";
     case MHDFStripHitVersion::V1_2: return os<<"1.2";
+    case MHDFStripHitVersion::V2_0: return os<<"2.0";
     case MHDFStripHitVersion::V2_2: return os<<"2.2";
     default: return os<<"Unknown";
   }
@@ -97,8 +98,8 @@ struct MHDFStripHit_V1_2 {
   uint8_t  m_CRC;
 };
 
-//! Version 2.2 of the HDF5 hit info
-struct MHDFStripHit_V2_2 {
+//! Version 2 of the HDF5 hit info
+struct MHDFStripHit_V2 {
   uint32_t m_EventIndex;
   uint8_t  m_HitType;
   uint8_t  m_TimingType;
@@ -107,7 +108,18 @@ struct MHDFStripHit_V2_2 {
   uint16_t m_TimingData;
 };
 
-//! Version 2.2 of the HDF5 event info
+//! Version 2.0 of the HDF5 event info
+struct MHDFEvent_V2_0 {
+  uint16_t m_EventID;
+  uint64_t m_TimeCode;
+  double   m_GSETimeCode;
+  uint8_t  m_Hits;
+  uint16_t m_Bytes;
+  uint8_t  m_EventType;
+  uint8_t  m_CRC;
+};
+
+//! Version 2.2 of the HDF5 event info (adding SPWTimeCode)
 struct MHDFEvent_V2_2 {
   uint16_t m_EventID;
   uint64_t m_TimeCode;
@@ -117,6 +129,17 @@ struct MHDFEvent_V2_2 {
   uint16_t m_Bytes;
   uint8_t  m_EventType;
   uint8_t  m_CRC;
+};
+
+//! HDF5 EventIndices (for HDF5v2 data)
+struct MHDFEventIndices_V2 {
+  uint32_t m_FEEHits[2];
+  uint32_t m_ACSHits[2];
+  uint32_t m_HS[2];
+  uint32_t m_SinglesCounts[2];
+  uint32_t m_DIBCoincidence[2];
+  uint32_t m_DetectorHits[2];
+  uint32_t m_DetectorLiveTime[2];
 };
 
 //! The version string
@@ -215,10 +238,16 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
   //! The HDF5 data set
   DataSet m_HDFDataSet;
   DataSet m_EventDataSet;
+  DataSet m_EventIndicesDataSet;
 
-  //! The HDF5 compond data type
+  //! The HDF5 compound data type (v1)
   CompType m_HDFCompoundDataType;
+
+  //! The HDF5 compound data types (v2)
   CompType m_EventCompoundDataType;
+  CompType m_EventIndicesCompoundDataType;
+  CompType m_HDFFEECompoundDataType;
+
 
   //! The version of the strip hit structure
   MHDFStripHitVersion m_HDFStripHitVersion;
@@ -231,14 +260,17 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
 
   //! The current index in the batch
   unsigned int m_CurrentBatchIndex;
+  unsigned int m_MinHitIndex;
 
   // The various batches:
   //! The MHDFStripHit_V1_0 batch:
   vector<MHDFStripHit_V1_0> m_Buffer_1_0;
   //! The MHDFStripHit_V1_2 batch:
   vector<MHDFStripHit_V1_2> m_Buffer_1_2;
-  //! The MHDFStripHit_V2_2 batch:
-  vector<MHDFStripHit_V2_2> m_Buffer_2_2;
+  //! The MHDFStripHit_V2 batch:
+  vector<MHDFStripHit_V2> m_Buffer_2;
+  vector<MHDFEventIndices_V2> m_EventIndices_2;
+  vector<MHDFEvent_V2_0> m_EventData_2_0;
   vector<MHDFEvent_V2_2> m_EventData_2_2;
 
   //! Total number of hits in a file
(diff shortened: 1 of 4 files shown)


zoglauer:
Thanks for starting this and taking it off my todo list, but there are a few more things to do:
(1) Based on info in the JSON ASIC setup file, we have to switch between strip maps - that is what Clio was supposed to implement in python and we just replicate it - but she thinks there is still a bug in her code - that is what we talked about on Tuesday.
(2) We need all the other info too: Live time, ACS hit info, etc.

fhagemann:
Thanks for the feedback, I will try to look into this.
Should this all become one big PR (this PR) and review all the code at the end?
Or do you want to tackle this in multiple PRs, each doing once change at a time, to keep reviews manageable?

zoglauer:
I don't have v2.3 but in previous 2.x versions, there was an EventIndices table "Maps each packet in Events to  a [start_index, end_index) range in each hit table that contains the hits from that event. An event at index i in the
Events table corresponds to the indices listed at index i in this table."
If it is still there in version 2.3, I don't see it in your code, to do the right linking

zoglauer:
I think we have to do (1) and the EventIndices if they are still in version 2.3 - because otherwise the output is not correct. And then we can do the rest of the content in a later PR

fhagemann:
With the latest commit, I seem to be able to read the polarities from all 3 primary/secondary asics for all 16 detectors in the `Config` attribute of `Events`:
```
Geometry materials loaded!
Detector ID 0:
  primary: HV HV HV 
  secondary: LV LV LV 
Detector ID 1:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 2:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 3:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 4:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 5:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 6:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 7:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 8:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 9:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 10:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 11:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 12:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 13:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 14:
  primary: LV LV LV 
  secondary: LV LV LV 
Detector ID 15:
  primary: LV LV LV 
  secondary: LV LV LV 
```

I still need to implement using this information for strip mapping.

fhagemann:
From my understanding: instead of "switching" between strip maps, the ASIC polarities are used to update if an ASIC is on HV or LV side, by ignoring that entry in the strip map and calculating it from 
- detector ID, 
- if it's the primary/secondary ASIC, and 
- the ASIC number
This is what `StripMap::UpdateASICPolarities` would now be doing.

fhagemann:
Ok, the code in this PR still "only" allows reading FEEHits, 
but it should be written in a way that is easy to extend to all other fields (ACSHits, H&S, DetectorLiveTime, ...)

There is a slight difference between v2.0-v2.1 and v2.2+, (v2.2+ added spacewire timing via `spw_timecode`), so I also split the v2 support into v2.0 and v2.2 support.

include/MModuleLoaderMeasurementsHDF.h
```diff
@@ -206,18 +260,25 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
...
   // The various batches:
   //! The MHDFStripHit_V1_0 batch:
   vector<MHDFStripHit_V1_0> m_Buffer_1_0;
   //! The MHDFStripHit_V1_2 batch:
   vector<MHDFStripHit_V1_2> m_Buffer_1_2;
+  //! The MHDFStripHit_V2 batch:
+  vector<MHDFStripHit_V2> m_Buffer_2;
+  vector<MHDFEventIndices_V2> m_EventIndices_2;
+  vector<MHDFEvent_V2_0> m_EventData_2_0;
+  vector<MHDFEvent_V2_2> m_EventData_2_2;
```
fhagemann, line 277: Two versions: v2.2+ allowing for spacewire timing (SPWTimeCode), while v2.0 doesn't

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -278,33 +424,79 @@ bool MModuleLoaderMeasurementsHDF::ReadBatchHits()
...
+      if (m_HDFStripHitVersion == MHDFStripHitVersion::V2_0) {
+        if (m_EventData_2_0.size() != Count[0]) {
+          m_EventData_2_0.resize(Count[0]);
+        }
+        m_EventDataSet.read(m_EventData_2_0.data(), m_EventCompoundDataType, MES, ES);
+      } else if (m_HDFStripHitVersion == MHDFStripHitVersion::V2_2) {
+        if (m_EventData_2_2.size() != Count[0]) {
+          m_EventData_2_2.resize(Count[0]);
+        }
+        m_EventDataSet.read(m_EventData_2_2.data(), m_EventCompoundDataType, MES, ES);
```
fhagemann, line 481: Is there a more elegant solution than defining a buffer for each format version?

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -376,74 +569,184 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
     // Extract the data we need
     uint16_t EventID;
     uint64_t TimeCode;
-    uint16_t StripID;
-    uint16_t ADCs;
-    uint16_t TACs;
+    uint64_t SPWTimeCode = 0;
```
fhagemann: Setting to 0, because only relevant in HDF5 v2.2+

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -376,74 +569,184 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
+
+        EventID = Hit.m_EventID;
+        TimeCode = Hit.m_TimeCode;
+        StripID = Hit.m_StripID;
+        ADCs = Hit.m_EnergyData;
+        TACs = Hit.m_TimingData;
+        NumberOfHits = Hit.m_Hits;
+        HitType = Hit.m_HitType;
+        TimingType = Hit.m_TimingType;
+      } else {
```
fhagemann, line 627: It was like this before, but can this somehow be combined?
Looks like a lot of code duplication to me..

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -376,74 +569,184 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
+        if (IsZeroDataBug == true) { // We are now out of the bug and need to recover
+          // Clear the strip hits - everything else gets overwritten later
+          while (Event->GetNStripHits() > 0) {
+            MStripHit* H = Event->GetStripHit(0);
+            delete H;
+            Event->RemoveStripHit(0);
+          }
+        }
+        IsZeroDataBug = false;
+      }
```
fhagemann, line 660: Is this something we would also expect in HDF v2?
The file structure has changed completely..

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -376,74 +569,184 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
+          // NOTE: at some point we will want to remove this code and always include nearest neighbor data
+          if (m_IncludeNearestNeighbor == false && Hit.m_HitType == 1) {
+            delete H; // Clean up the memory we just allocated
+            // Increase counters
+            NStripHits = static_cast<unsigned int>(NumberOfHits);
+            StripHitIndex++;
+            continue;
+          }
+            
+          Event->AddStripHit(H);
```
fhagemann, line 758: This is also duplicated right now (from later in the file).

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -456,51 +759,23 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
-      // If the user does not want to include Nearest Neighbors in the data, then this is where we remove them
-      // NOTE: at some point we will want to remove this code and always include nearest neighbor data
-      if (m_IncludeNearestNeighbor == false && HitType == 1) {
-        delete H; // Clean up the memory we just allocated
-        // Increase counters
-        NStripHits = static_cast<unsigned int>(NumberOfHits);
-        StripHitIndex++;
-        continue;
+      if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_2) {
+        Event->SetCL(SPWTimeCode);
```
fhagemann, line 782: I'm right now setting the `TimeCode` (which is `GSETimeCode` if existent) to `TI` and the `SPWTimeCode` to `CL`.
Is this how it should be?

fhagemann:
I tried parsing files in v1.1, v2.1 and v2.3, and I'm not getting any errors at least.
I asked @robin4730 (huge thanks!!) to look into if after the updates, the ROA files still look reasonable/identical.

src/MModuleLoaderMeasurementsHDF.cxx
zoglauer: The strip hit definitions are different and not derivable from each other. Thus this is the only thing we can do...

src/MModuleLoaderMeasurementsHDF.cxx
zoglauer: Unfortunately, the buffers contain different objects, thus this is the only way to do it (without other quirks which are even more horrible)

src/MModuleLoaderMeasurementsHDF.cxx
zoglauer: Make it a code comment!

src/MModuleLoaderMeasurementsHDF.cxx
zoglauer: I know, because the hits are different objects, and since the data needs to be aligned in a fixed way to extract it from HDF5, as far as I understand it, we cannot simplify it by deriving classes.

src/MModuleLoaderMeasurementsHDF.cxx
zoglauer: I don't know what it actually means.

Those are the times we currently have (Carolyn want to scrub it eventually):
  //! TODO Scrub all clock/time variables for COSI SMEX 
  //! Set the Frame Counter of this event
  void SetFC(unsigned int FC) { m_FC = FC; }
  //! Return the Frame Counter of this event
  unsigned int GetFC() const { return m_FC; }

  //! set and get Unix clock time
  void SetTI(unsigned long long TI) { m_TI = TI;}
  unsigned long long GetTI() const { return m_TI;}

  //! set and get clock tick
  void SetCL(uint64_t CL) { m_CL = CL;}
  uint64_t GetCL() const { return m_CL;}

  //! Set and get the Modified Julian Date of this event
  void SetMJD(double MJD) { m_MJD = MJD; }
  double GetMJD() const { return m_MJD; }
  
  //! Set and get the UTC time of this event
  void SetTimeUTC(const MTime& TimeUTC) { m_EventTimeUTC = TimeUTC; }
  MTime GetTimeUTC() const { return m_EventTimeUTC; }

src/MStripMap.cxx
```diff
@@ -101,6 +101,18 @@ bool MStripMap::Open(MString FileName)
 }
 
 
+////////////////////////////////////////////////////////////////////////////////
+
+bool MStripMap::UpdateASICPolarities(vector<map<bool, vector<bool>>> ASICPolarities) {
+  if (!m_StripMappings.empty()) {
+    for (MSingleStripMapping &s : m_StripMappings) {
```
zoglauer: MSingleStripMapping& S

include/MModuleLoaderMeasurementsHDF.h
```diff
@@ -206,18 +260,25 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
...
 
   //! Total number of hits in a file
   hsize_t m_TotalHits;
 
   //! Current hit number in the file
   hsize_t m_CurrentHit;
+  unsigned int m_CurrentEvent;
```
zoglauer: Why is this not hsize_t? hsize_t is unsigned long, thus we should use that too

zoglauer: And where is it used?

src/MModuleLoaderMeasurementsHDF.cxx
zoglauer: SPWTimeCode is double in the definition - can you check it again?

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -376,74 +569,184 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
-      TACs = Hit.m_TimingData;
-      NumberOfHits = Hit.m_Hits;
-      HitType = Hit.m_HitType;
-      TimingType = Hit.m_TimingType;
-    
-    } else {
-      if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Unhandled HDF hit version found: "<<m_HDFStripHitVersion<<endl<<"Please update this module."<<endl;
-      return false;
-    }
+      if (m_Buffer_2.data() == nullptr || m_Buffer_2.size() == 0) {
```
zoglauer: Can you check if m_Buffer_2.data() can ever be a nullptr? I think it cannot.

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -240,12 +334,64 @@ bool MModuleLoaderMeasurementsHDF::OpenHDF5File(MString FileName)
...
+
+      // Create compound data type for reading event indices information from /EventIndices
+      hsize_t array_dims[1] = {2};
+      ArrayType uint32_pair(PredType::STD_U32LE, 1, array_dims);
+      m_EventIndicesCompoundDataType = CompType(sizeof(MHDFEventIndices_V2));
+      m_EventIndicesCompoundDataType.insertMember("fee_hits",            HOFFSET(MHDFEventIndices_V2, m_FEEHits),            uint32_pair);
+      m_EventIndicesCompoundDataType.insertMember("acs_hits",            HOFFSET(MHDFEventIndices_V2, m_ACSHits),            uint32_pair);
+      m_EventIndicesCompoundDataType.insertMember("hs",                  HOFFSET(MHDFEventIndices_V2, m_HS),                 uint32_pair);
+      m_EventIndicesCompoundDataType.insertMember("singles_counts",      HOFFSET(MHDFEventIndices_V2, m_SinglesCounts),      uint32_pair);
+      m_EventIndicesCompoundDataType.insertMember("dib_coindicidence",   HOFFSET(MHDFEventIndices_V2, m_DIBCoincidence),     uint32_pair);
```
zoglauer: dib_coindicidence - is this an intentional typo?

zoglauer:
Generally looks good, just a bunch of small things.

include/MModuleLoaderMeasurementsHDF.h
fhagemann: This is still a left-over from the first implementation and is not used anymore.
I will remove it.

src/MModuleLoaderMeasurementsHDF.cxx
fhagemann: Good catch, this was unintentional.
I will fix it.

src/MModuleLoaderMeasurementsHDF.cxx
fhagemann: Good catch, the defintion is wrong. `GSETimeCode` used to be `double` in HDF version 1.2, but changed to `uint64_t` in HDF v2 (ns since epoch), and `SPWTimeCode` is also `uint64_t` (ns since start of acquisition).
Image

I will update the definition and add comments to the definition.

src/MStripMap.cxx
fhagemann: Done!

src/MModuleLoaderMeasurementsHDF.cxx
fhagemann: I didn't touch that part of the code with my latest commit, but this goes along with #69.
We should eventually decide how/where we want to store that information (also for DEE output).

src/MModuleLoaderMeasurementsHDF.cxx
fhagemann: Yeah, this was redundant.. I changed the `if` statement to `m_Buffer_2.empty()`

zoglauer:
Put you name into the copyright notices!

fhagemann:
Done (I also added my name to the copyright notices of `MStripMap` -- even though the changes in this PR might be minor, there were some bigger changes I made in #103)

zoglauer:
Can you check if you initialize or clear everything in the Initialize function?
I don't see m_ASICPolarities there

fhagemann:
That should also be done now. The only thing I don't initialize is `m_HDFStripHitVersion` (this is set at the beginning of `OpenHDF5File`, but I could also move it to `Initialize` to have everything initialized there)

And `m_StripMap` is also handled by `m_StripMap::Open`, but I could also explicitly clear it in `Initialize` if you want.

zoglauer:
OK. Feel free to merge.

PR #101, opened 2026-03-03
DEE Shield energy correction with new naming

vfioretti:
- DEE Shield energy correction with DetectorID as string (e.g. X0). 
- Standard error/warning/info handling with verbosity
- Minor bug correction in the shield energy correction class: if the new energy is below 0 is set to 0.
- New energy correction file in nuclearizer/resource/dee/ named ShieldEnergyCorrection_v2.csv


ckierans:
Is this ready to review?

vfioretti:
Yes

zoglauer:
I don't see anything obvious issues. But could you check how closely you follow the style guide in these classes:
https://github.com/cositools/nuclearizer/blob/develop/em/CodingConventions.md

resource/dee/ShieldEnergyCorrection_v2.csv
```diff
@@ -0,0 +1,1374 @@
...
+X0         0          0         0        10    -10.21    -54.08      0.97      4.57     18.89      2.88      0.00
+X0         0          0         0         9      0.00    -54.08      1.00      1.55     27.46      2.48      0.00
+X0         0          0         0         8     10.21    -54.08      0.98      3.14     22.61      2.63      0.00
+X0         0          0         0         7     20.42    -54.08      0.95      3.81     16.63      2.72      0.00
+X0         0          0         0         6     30.63    -54.08      0.95      3.41     24.89      2.66      0.00
+X0         0          0         0         5     40.84    -54.08      0.94      2.12     24.16      2.35      0.05
+X0         0          0         0         4     51.05    -54.08      0.95      1.83     25.30      2.42      0.04
+X0         0          0         0         3     61.26    -54.08      0.95      1.71     23.72      2.65      0.00
+X0         0          0         0         2     71.47    -54.08      0.95      1.80     19.55      2.81      0.00
+X0         0          0         0         1     81.68    -54.08      0.93      4.14     24.07      2.46      0.05
```
ckierans, line 24: For this file format, it's strange that the X and Y voxels numbers increase in each row, but the voxel_Z decreases (starts at 18 then goes to 0). It's probably annoying to swap, but can you be consistent?

Is there a reason you don't record the z position in mm?

Also, just making sure I understand, do these voxel numbers align with the mass model voxelization in the BGO crystals, or is that completely seperate?

src/MSubModuleShieldEnergyCorrection.cxx
```diff
@@ -96,39 +96,77 @@ void MSubModuleShieldEnergyCorrection::Clear()
...
+  // Energy correction
+  list<MDEECrystalHit>& Hits = Event->GetDEECrystalHitListReference();
+  for (MDEECrystalHit& CH : Hits) {
+
+     const double true_energy = CH.m_SimulatedEnergy;
+      
+     if (true_energy < 0) {
+        if (g_Verbosity >= c_Warning) cout << "WARNING: Negative simulated energy (" << true_energy
+              << ")" << endl;
+        CH.m_Energy = true_energy;
```
ckierans, line 117: The simulated hits haven't been modified in the DEE yet, so why would you ever expect a negative simulated energy? And if a negative energy is found, why do we just skip the event here? Should we instead set the energy to zero and smear it?

src/MSubModuleShieldEnergyCorrection.cxx
```diff
@@ -96,39 +96,77 @@ void MSubModuleShieldEnergyCorrection::Clear()
...
+     
+     MString det_id = CH.m_DetectorID;
+     int crystal_id = CH.m_CrystalID;
+
+     MVector voxel_id = CH.m_VoxelInDetector;
+
+     const double shield_corrected_centroid = NoiseShieldEnergyCentroid(true_energy, det_id, crystal_id, voxel_id[0], voxel_id[1], voxel_id[2]);
+     const double shield_FWHM_value = NoiseShieldEnergyFWHM(true_energy, det_id, crystal_id, voxel_id[0], voxel_id[1], voxel_id[2]);
+
+     // If FWHM not available or invalid, do not smear
```
ckierans, line 129: Why not? When do you expect this to happen and why would we not have a default smear or something? I'm not saying this is wrong, I'm just trying to understand.

src/MSubModuleShieldEnergyCorrection.cxx
```diff
@@ -96,39 +96,77 @@ void MSubModuleShieldEnergyCorrection::Clear()
...
+        continue;
+     }
+      
+     double shield_sigma = shield_FWHM_value / 2.35;
+     double corrected_energy = m_Random.Gaus(shield_corrected_centroid, shield_sigma);
+
+     if (corrected_energy < 0) {
+         if (g_Verbosity >= c_Warning) cout  << "WARNING: Corrected energy negative (" << corrected_energy
+              << ") from centroid = " << shield_corrected_centroid << " sigma=" << shield_sigma
+              << " -> clamping to 0" << endl;
```
ckierans, line 143: "clamping to 0" -> "setting to 0 keV"

ckierans (commented):
Looks good, Valentina. I have a few questions above but they're mostly for my understanding.

resource/dee/ShieldEnergyCorrection_v2.csv
vfioretti: Thanks for taking the time to review this!
My replies:
- For the voxel_Z order, yes you're right. I am organizing Alex's code and I will change that.
- The position is not used in the generation of the energy corrections, and it has a different system (X and Y are the sides of the crystal, Z the thickness) than the voxel_IDs. I kept X and Y in the file because they could be useful in the verification phase, and there is no Z because the energy correction is symmetric along the Z axis (the 2 cm thickness).
- yes, the voxel_IDs are aligned with the ones of the mass model, so that I read from the .sim the voxel_ID and from the csv file they are mapped to the correspondent parameters. Said that, without a .tra that also records the energy in the BGO, there is no way for me to verify that the DEE is working as it should.

src/MSubModuleShieldEnergyCorrection.cxx
vfioretti: It shouldn't happen, but it could be that some of the parameters are corrupted. I am working on an automatic pipeline that generates them from fitting, and with many crystals and voxels could be that one of the parameters is corrupted. If that happens, as a developer I would then see it in the warning and correct the issue, but for the user that is using that particular file, the DEE will go on without blocking nuclearizer.

Your are right, if this happens better to use a default value, I'll correct that.

src/MSubModuleShieldEnergyCorrection.cxx
vfioretti: Ok, I'll change that.

vfioretti:
> I don't see anything obvious issues. But could you check how closely you follow the style guide in these classes: https://github.com/cositools/nuclearizer/blob/develop/em/CodingConventions.md

Thanks @zoglauer. I didn't know they existed. I'll also check this and upload a new pull request with @ckierans 's comments integrated.

src/MSubModuleShieldEnergyCorrection.cxx
vfioretti: I may have added too many warnings...
For this part, there should be no negative energy. If that happens something really wrong is occurring before (for example when building the crystal hit). Looking at this again, I should add an error, not a warning.

vfioretti:
I am closing this pull request. I will upload a new pull request with the code compliant to the coding rules and comments  included.

PR #102, opened 2026-03-05
Add MGUIExpoPlotSpectrum for plotting energy spectrums

robin4730:
This update adds a new plotting tool, MGUIExpoPlotSpectrum, to plot the spectrum as it moves through the pipeline. The function to plot the spectrum using MGUIExpoPlotSpectrum can be called in any module. This is what the new plotting looks like (comparing before and after the TAC cut in this example): 

Screenshot 2026-03-04 at 9 11 07 PM

Key Features:
* Can compares 'Initial' and 'Final' energy distributions side-by-side to verify that cuts are working as expected.
* Separates LV and HV hits
* Plots Strips and Nearest Neighbor data separately 
* Once added to a module, the user can choose whether or not to plot the data. There are three options:   
    - 'No Plot', the data is not plotted. This is best for running from the command line. 
    - 'No Buffer' for a live look at the spectrum during a run. 
    - 'With Buffer' if you have the buffer all the data is saved as a copy so if you changed something like the bin size, all the data is reloaded from the buffer and plotted. This means you can look at the data after the run finishes. But it also means that x2 the memory is used. 

This is what the plotting options look like (added to the TAC cut GUI option): 

Screenshot 2026-03-04 at 9 04 06 PM

The tool is integrated into the Energy Calibration Universal and TAC Calibration modules, with all plotting preferences and modes saved directly to the XML configuration files.

### DIFF
--- a/include/MGUIExpoPlotSpectrum.h
+++ b/include/MGUIExpoPlotSpectrum.h
@@ -65,6 +65,9 @@ public:
 
   //! Update the frame
   virtual void Update();
+  
+  //! Set the plotting mode (200=None, 201=NoBuffer, 202=WithBuffer)
+  void SetPlotMode(int Mode) { m_PlotSpectrumMode = Mode; }
 
 
 protected:
@@ -111,7 +114,7 @@ private:
   std::vector<double> m_DataBufferNearestNeighborLVFinal;
   std::vector<double> m_DataBufferNearestNeighborHVFinal;
   
-  bool m_PlotSpectrum;
+  int m_PlotSpectrumMode;
 
 #ifdef ___CLING___
 public:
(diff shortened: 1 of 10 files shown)


zoglauer:
Hi Robin,

Instead of adding it to a module - which adds some duplicated code, have you thought about making this its own module? Then the user can plot anywhere they want.

robin4730:
I did, and that was my initial goal. But I could not figure out how to access "initial" and "final" data from a module. I think it would requite being able to access the data from the datastream before it enters a module and after it leaves that module. Is that possible?

zoglauer:
Let me know when you are in next and we can talk it through - I am free until 3pm

zoglauer:
Before I do a full review, can you fix your multiple plot modes. I don't like that we have to use the numbers so often.

You have:
enum MEnergyCalibrationPlotSpectrumModes : int
{
  e_PlotNone = 200,
  e_PlotNoBuffer = 201,
  e_PlotWithBuffer = 202
};

Instead, do in the header of MGUIExpoPlotSpectrum.h‎

//! Definition of the plotting modes
enum class MGUIExpoPlotSpectrumPlotModes : int { 
  c_None = 200, c_NoBuffer = 201, c_WithBuffer = 202
};

Then you just define it once in the plot class, and use it in the classes which use the plot module, e.g.

void SetPlotSpectrumMode(const MGUIExpoPlotSpectrumPlotModes& Mode) { m_PlotSpectrumMode = Mode; }

This should make the class a bit easier to use, and you don;t nee to redefine the enum

And then don't so any number:
if (m_PlotSpectrumMode == 202) {
-->
if (m_PlotSpectrumMode == MGUIExpoPlotSpectrumPlotModes:: c_WithBuffer) {


If you fix that, then I check the rest.

PR #114, opened 2026-03-23
Added GR and shield veto functionality

### LINKED ISSUE #95: Guard ring strip number
Strip 64 will currently give an error in the StripTrigger module. This is supposed to be the strip number for guard ring.

Added this issue so we can keep track of it.

parshadkp:
GR and shield veto added. Checked response correctness by using a low energy and a high energy Cosima source. May need to update m_ShieldVetoWindowSize in the future.

This will close issue: #95.


src/MSubModuleDEEIntake.cxx
```diff
@@ -132,7 +132,7 @@ bool MSubModuleDEEIntake::AnalyzeEvent(MReadOutAssembly* Event)
     double DetectorDepth = Shape->GetSizeZ();
 
     MString DetectorName = Detector->GetName();
-    if (DetectorName.BeginsWith("GeD") == true) {
+    if (DetectorName.BeginsWith("GeD") == true || DetectorName.BeginsWith("GuardRing")) {
       DetectorName.RemoveAllInPlace("GeD_"); // The number after GeD is the COSI detector ID
```
fhagemann: Does this work? I believe you might also need to update the code underneath the `if`-statement to retrieve the correct `DetectorID`.

fhagemann: This might work, but also only if the string following `GeD_` or `GuardRingDetector_GeD_` is a number
(which seems like is the case in the `AutoDetectorBuild` branch of `massmodel-cosi-em`)
```suggestion
    if (DetectorName.BeginsWith("GeD") == true || DetectorName.BeginsWith("GuardRing")) {
      DetectorName.RemoveAllInPlace("GuardRingDetector_"); // Remove prefix GuardRing if existent
      DetectorName.RemoveAllInPlace("GeD_"); // The number after GeD is the COSI detector ID
```

parshadkp:
> GR and shield veto added. Checked response correctness by using a low energy and a high energy Cosima source. May need to update m_ShieldVetoWindowSize in the future.
> 
> This will close issue: #95.

Updated code to read m_ShieldVetoWindowSize from deadtime_Parameters.txt file, which can be easily changed. Also addressed Felix's comment in latest commit.

src/MSubModuleDEEIntake.cxx
```diff
@@ -132,7 +132,8 @@ bool MSubModuleDEEIntake::AnalyzeEvent(MReadOutAssembly* Event)
     double DetectorDepth = Shape->GetSizeZ();
 
     MString DetectorName = Detector->GetName();
-    if (DetectorName.BeginsWith("GeD") == true) {
+    if (DetectorName.BeginsWith("GeD") == true || DetectorName.BeginsWith("GuardRing")) {
```
fhagemann:
```suggestion
    if (DetectorName.BeginsWith("GeD") == true || DetectorName.BeginsWith("GuardRing") == true) {
```
I've seen explicit `== true` statement elsswhere (?)

include/MSubModuleStripTrigger.h
```diff
@@ -92,7 +92,7 @@ class MSubModuleStripTrigger : public MSubModule
   double GetStripsTotalDeadtime() const { return m_StripsTotalDeadtime; }
 
   //! Get number of strip hits erased
-  int GetStripHitsErased() const { return m_StripHitsErased; }
+  // int GetStripHitsErased() const { return m_StripHitsErased; }
```
fhagemann, line 95: Can you comment on why you removed `m_StripHitsErased` and all associated functionalities? Will this be dealt with in another (sub)module?

src/MSubModuleShieldTrigger.cxx
```diff
@@ -173,24 +171,21 @@ double MSubModuleShieldTrigger::CalculateASICDeadtime(vector<int> CrystalIDs)
...
-    }
-  }
+  // // Track which GeD detectors got hit (for later deadtime update)
+  // list<MDEEStripHit>& LVHits = Event->GetDEEStripHitLVListReference();
+  // for (const MDEEStripHit& Hit : LVHits) {
+  //   int DetID = Hit.m_ROE.GetDetectorID();
+  //   if (DetID >= 0 && DetID < nDets) {
+  //     m_DetectorsHitForShieldVeto[DetID] = 1;
+  //   }
+  // }
```
fhagemann, line 183: Why was this commented? Do you plan on adding it back later (in a modified form)?

parshadkp: Nope, I will remove this permanently. This was leftover from old balloon code.

include/MSubModuleStripTrigger.h
parshadkp: No other modules uses this and this was only used for the printing at the end of the StripTriggerModule. This was just meant to be a diagnostic printout.

src/MSubModuleDEEIntake.cxx
```diff
@@ -132,7 +132,8 @@ bool MSubModuleDEEIntake::AnalyzeEvent(MReadOutAssembly* Event)
     double DetectorDepth = Shape->GetSizeZ();
 
     MString DetectorName = Detector->GetName();
-    if (DetectorName.BeginsWith("GeD") == true) {
+    if (DetectorName.BeginsWith("GeD") == true || DetectorName.BeginsWith("GuardRing") == true) {
+      DetectorName.RemoveAllInPlace("GuardRingDetector_GeD_"); // Remove prefix GuardRing if existent
```
ckierans, line 136: Just so I understand. We're checking for GuardRing here because the Geometry volume we're using in the DEE has the guard ring and the GeD as separate volumes. But the simulation has them all as one solid volume. Excluding this "GuardRingDetector_GeD_" here would exclude any hits that are within the GR volume. But we're not trying to distinguish Guardring hits here, that only comes after the charge transport, so if there is a simulated hit that is within the GuardRing volume in the Nuclearizer mass model, then it gets added as a regular hit in the detector # that precedes GuardRingDetector_GeD_. Right?

src/MModuleDEESMEX.cxx
```diff
@@ -192,8 +193,7 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
     if (m_StripTrigger.GetDeadTimeEnd() > m_DeadTimeEnd) {
       m_DeadTimeEnd = m_StripTrigger.GetDeadTimeEnd();
     }
-    // Clean up
-
+    Event->SetGuardRingVeto(true);   // <-- mark the event so EventSaver can filter it
```
ckierans, line 201: It's not clear to me that this "m_StripTrigger.HasVeto()" corresponds to a GuardRingVeto vs a Shield Veto. Can we improve the variable names to clarify it's a shield veto or GR veto. "HasVeto" should encompass both veto cases, and we should probably introduce a HasShieldVeto and HasGuardRingVeto variable. See, for example, the veto variable names in MReadOutAssembly:

src/MSubModuleShieldTrigger.cxx
```diff
@@ -65,11 +65,9 @@ MSubModuleShieldTrigger::MSubModuleShieldTrigger() : MSubModule()
   m_ShieldDelayBefore = 0.1e-6;
   m_ShieldDelayAfter = 0.4e-6;
   m_ShieldVetoWindowSize = 1.5e-6;
-  m_ASICDeadTimePerChannel = 0.0;
-  m_ShieldVetoTime = 0.0;
+  m_ASICDeadTimePerChannel = 1e-6;
```
ckierans: Shouldn't these variables come from the ParseDeadtimeFiles function? Is this just a placeholder for now?

resource/calibration/COSI_SingleDet/deadtimeParameters.txt
```diff
@@ -1,2 +1,4 @@
-CoincidenceWindow (s); DeadtimePerStrip (s); delayAfter1 (s); delayAfter2 (s)
-1.4e-6		       1e-6	             0.2e-6	      0.2e-6
\ No newline at end of file
+CoincidenceWindow (s); StripDeadtimePerStrip (s); delayAfter1 (s); delayAfter2 (s)
+1.4e-6		       1e-6	                  0.2e-6	   0.2e-6
+ShieldDeadtimePerStrip (s); ShieldVetoWindow (s); ShieldPulseDuration (s); ShieldDelayBefore (s); ShieldDelayAfter (s)
+1e-6                        1.5e-6                1.7e-6                   0.1e-6                 0.4e-6
```
ckierans, line 4: Do we expect any of these parameters to change detector by ASIC by ASIC or anything? Or are these parameters which will be valid for all GeDs and all shields?

src/MSubModuleShieldTrigger.cxx
```diff
@@ -65,11 +65,9 @@ MSubModuleShieldTrigger::MSubModuleShieldTrigger() : MSubModule()
   m_ShieldDelayBefore = 0.1e-6;
   m_ShieldDelayAfter = 0.4e-6;
   m_ShieldVetoWindowSize = 1.5e-6;
-  m_ASICDeadTimePerChannel = 0.0;
-  m_ShieldVetoTime = 0.0;
```
ckierans, line 69: Why do we no longer need m_ShieldVetoTime? Is this redundant with m_ShieldVetoWindowSize or m_ASICDeadTimePerChannel?

include/MSubModuleShieldTrigger.h
```diff
@@ -97,9 +97,6 @@ class MSubModuleShieldTrigger : public MSubModule
   //! Get shield hit counts
   int GetShieldHitCounts() const { return m_NumShieldHitCounts; }
 
-  //! Get shield veto counter
-  int GetShieldVetoCounter() const { return m_NumShieldVetoCounts; }
```
ckierans, line 101: Do we not want to track the number of vetos with this counter?

ckierans (commented):
Thanks, Parshad! Things look good, but I have some questions. The big comment is that i think we want to separate out the HasGuardRingVeto and HasShieldVeto variables to better track what's happening and make the code easier to read.

src/MSubModuleDEEIntake.cxx
parshadkp: Correct, the charge transport module will assign this hit a Strip ID of 64. Here we are just assigning the detector ID.

src/MSubModuleShieldTrigger.cxx
parshadkp: m_ShieldVetoWindowSize is exactly m_ShieldVetoTime. I think this was leftover from the balloon.

src/MSubModuleShieldTrigger.cxx
parshadkp: They do, these are just the default values. I can remove these if you think it's more suitable.

include/MSubModuleShieldTrigger.h
parshadkp: This used to return a really high number. I think the logic was incorrect. I was not using the number of vetoes anywhere other than a diagnostic check. Is this number important to have?

src/MModuleDEESMEX.cxx
parshadkp: GR veto comes from m_StripTrigger.HasVeto(). Shield veto comes from m_ShieldTrigger.HasVeto(). Are you suggesting to change these to maybe "m_StripTrigger.HasGRVeto()" and "m_ShieldTrigger.HasShieldVeto()"?

resource/calibration/COSI_SingleDet/deadtimeParameters.txt
parshadkp: These will be constant for all GeDs and all shields.

include/MSubModuleShieldTrigger.h
ckierans: Diagnostic checks are helpful, but if you're not using it then and/or if it's returning the wrong number, then it's okay to remove. We can always add something back in when doing checks at a later time, if we think it's necessary.

src/MSubModuleShieldTrigger.cxx
ckierans: I think having the default values written here is a bit confusing since they'll just be rewritten. When someone is looking at the code quickly someone might think editing these numbers will change the processing, but that's not true. I'd remove them here for clarity.

src/MModuleDEESMEX.cxx
ckierans: Hmm, I'm conflicted. It makes sense for the GR veto to only apply to the StripTrigger and  shield veto to only apply to ShieldTrigger when you think about this low-level of processing, but it's not clear this is the case based on the variable names or the comments in the header files.

At a minimum: let's define the specific veto type in the header file comments. For example, for the MSubModuleStripTrigger.h file:

>   //! Return true if we have a Guard Ring veto - filled after AnalyzeEvent
>   bool HasVeto() const { return m_HasVeto; }

And the same for the ShieldTrigger header file. Furthermore, adding comments in the code when you define the HasVeto to make it clear that it's GR or Shield would help the readability.

I still think changing the variable names will make the code easier to understand. But with sufficient comments in the code I think leaving it should be okay.

ckierans (commented):
Thanks, Parshad. Sorry for the delay.

src/MModuleDEESMEX.cxx
parshadkp: Changed the variable names to HasGRVeto() and HasShieldVeto() and added comments in .h files.

src/MSubModuleShieldTrigger.cxx
parshadkp: Removed the default values.

ckierans:
Okay, thanks for the changes, @parshadkp! Things look good!

PR #116, opened 2026-03-24
Fixing GR Hit QA message bug

julianmgerber:
Fixing bug that duplicated GR Hit info in QA flag


ckierans (commented):
Thanks for the quick work!

PR #117, opened 2026-03-25
[DEE] Implement inverse energy calibration

fhagemann:
I created this PR too early, so someone might need to redirect it from `main` to `develop/em` (I can't seem to do that myself).

The credit for implementing the inverse energy calibration for DEE should all go to @robin4730, the only thing I added to her nice implementation was to remove code duplication, restrict the determination of ADCs from energies to the ADC range, and to remove the old dummy implementation for the inverse energy calibration that would otherwise overrule what's happening here.


zoglauer:
Hi Felix,
Just make a new pull request....

fhagemann:
I will, sorry for the noise...

PR #118, opened 2026-03-25
[DEE] Implement inverse energy calibration

fhagemann:
The credit for implementing the inverse energy calibration for DEE should all go to @robin4730.

The only thing I added to her nice implementation was:
- removing code duplication, 
- restricting the determination of ADCs from energies to the ADC range, and
- remove the old dummy implementation for the inverse energy calibration 
  (which would otherwise overrule what's happening here)

### DIFF
--- a/include/MSubModuleStripReadout.h
+++ b/include/MSubModuleStripReadout.h
@@ -17,9 +17,9 @@
 
 
 // Standard libs:
+#include <map>
 
 // ROOT libs:
-#include <map>
 #include "TF1.h"
 
 // MEGAlib libs:
--- a/include/MSubModuleStripReadout.h
+++ b/include/MSubModuleStripReadout.h
@@ -19,10 +19,13 @@
 // Standard libs:
 
 // ROOT libs:
+#include <map>
+#include "TF1.h"
 
 // MEGAlib libs:
 #include "MGlobal.h"
 #include "MSubModule.h"
+#include "MReadOutElementDoubleStrip.h"
 
 // Forward declarations:
 
@@ -87,6 +90,12 @@ class MSubModuleStripReadout : public MSubModule
  private:
   //! Energy calibration file name
   MString m_EnergyCalibrationFileName;
+  
+  //! Name of the strip map
+  std::map<MReadOutElementDoubleStrip, TF1*> m_Calibration;
+  
+  //! Make value of the ADC units
+  double m_MaxADCRange;
 
 
 
--- a/src/MSubModuleStripReadout.cxx
+++ b/src/MSubModuleStripReadout.cxx
@@ -185,51 +185,29 @@ bool MSubModuleStripReadout::AnalyzeEvent(MReadOutAssembly* Event)
 {
   // Main data analysis routine, which updates the event to a new level
 
-  // Get low voltage hits
-  list<MDEEStripHit>& LVHits = Event->GetDEEStripHitLVListReference();
-  for (MDEEStripHit& SH: LVHits) {
+  // Get low-voltage and high-voltage hits
+  for (auto* Hits : { &Event->GetDEEStripHitLVListReference(), &Event->GetDEEStripHitHVListReference() }) {
     
-    // Look up the fit using the ecal
-    TF1* Fit = m_Calibration[SH.m_ROE];
-
-    if (Fit != nullptr) {
-      // Run it in reverse, using ROOT's poly inverter (keV -> ADC)
-      double calculatedADC = Fit->GetX(SH.m_Energy);
-      
-      // Apply hardware limits
-      if (calculatedADC > m_MaxADCRange) calculatedADC = m_MaxADCRange;
-      if (calculatedADC < 0) calculatedADC = 0;
-      
-      SH.m_ADC = static_cast<unsigned int>(calculatedADC);
-      
-    } else {
-      // If no calibration exists in the .ecal file for this strip set it to ADC value of 0
-      if (g_Verbosity >= c_Warning) cout << m_Name << ": No inverse calibration found for LV element " << SH.m_ROE << endl;
-      SH.m_ADC = 0;
-    }
-  }
-
-  // Now do high voltage strips
-  list<MDEEStripHit>& HVHits = Event->GetDEEStripHitHVListReference();
-  for (MDEEStripHit& SH: HVHits) {
+    for (MDEEStripHit& SH : *Hits) {
     
-    // Find the fit
-    TF1* Fit = m_Calibration[SH.m_ROE];
+      // Look up the fit using the ecal
+      TF1* Fit = m_Calibration[SH.m_ROE];
 
-    if (Fit != nullptr) {
-      // keV -> ADC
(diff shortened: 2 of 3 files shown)


include/MSubModuleStripReadout.h
```diff
@@ -19,10 +19,13 @@
 // Standard libs:
 
 // ROOT libs:
+#include <map>
```
zoglauer: map goes up to standard Lins

src/MSubModuleStripReadout.cxx
```diff
@@ -29,9 +29,13 @@
 // Standard libs:
 
 // ROOT libs:
+#include "TF1.h"
```
zoglauer: You included those 2 already in the header!

src/MSubModuleStripReadout.cxx
```diff
@@ -29,9 +29,13 @@
...
+#include <map>
 
 // MEGAlib libs:
 #include "MSubModule.h"
+#include "MParser.h"
+#include "MReadOutElementDoubleStrip.h"
```
zoglauer: Same here: What's in the header you don't need to include in the cxx

src/MSubModuleStripReadout.cxx
```diff
@@ -71,13 +78,99 @@ bool MSubModuleStripReadout::Initialize()
...
 
+  // Check if we have a file
+  if (m_EnergyCalibrationFileName == "") {
+    if (g_Verbosity >= c_Error) {
+      cout << m_Name << ": No energy calibration file specified." << endl;
+    }
+    return false;
+  }
+
+  // Opoen ecal file
```
zoglauer: Open

src/MSubModuleStripReadout.cxx
```diff
@@ -71,13 +78,99 @@ bool MSubModuleStripReadout::Initialize()
...
+  MParser Parser;
+  if (Parser.Open(m_EnergyCalibrationFileName, MFile::c_Read) == false) {
+    if (g_Verbosity >= c_Error) {
+      cout << m_Name << ": Unable to open calibration file " << m_EnergyCalibrationFileName << endl;
+    }
+    return false;
+  }
+
+  // Create the map (smae as the Universal Energy Calibrator)
```
zoglauer: same

src/MSubModuleStripReadout.cxx
```diff
@@ -71,13 +78,99 @@ bool MSubModuleStripReadout::Initialize()
...
+  map<MReadOutElementDoubleStrip, unsigned int> CM_ROEToLine;
+
+  // Add case for double wide strips
```
zoglauer: I don't see how you are handling double-wide strips - I assume you mean neighboring strips which are connected together?

src/MSubModuleStripReadout.cxx
```diff
@@ -71,13 +78,99 @@ bool MSubModuleStripReadout::Initialize()
...
+
+      TF1* melinatorfit = new TF1("poly3", "[0] + [1]*x + [2]*x^2 + [3]*x^3", 0., m_MaxADCRange);
+      melinatorfit->FixParameter(0, a0);
+      melinatorfit->FixParameter(1, a1);
+      melinatorfit->FixParameter(2, a2);
+      melinatorfit->FixParameter(3, a3);
+
+      m_Calibration[CM.first] = melinatorfit;
+    }
+    //TODO: @RobinAnthonyPetersen add all the other types of fits melinator can do
```
zoglauer: We should at least throw a mean error message. And maybe even return false.

src/MSubModuleStripReadoutNoise.cxx
```diff
@@ -91,6 +91,9 @@ bool MSubModuleStripReadoutNoise::AnalyzeEvent(MReadOutAssembly* Event)
 {
   // Main data analysis routine, which updates the event to a new level 
 
+  // TODO: Update this to take energy resolution into account
+  
+  /*
```
zoglauer, line 96: Do we still need this class?

src/MSubModuleStripReadout.cxx
```diff
@@ -71,13 +78,99 @@ bool MSubModuleStripReadout::Initialize()
...
+      R.SetDetectorID(Parser.GetTokenizerAt(i)->GetTokenAtAsUnsignedInt(2));
+      R.SetStripID(Parser.GetTokenizerAt(i)->GetTokenAtAsUnsignedInt(3));
+      R.IsLowVoltageStrip((Parser.GetTokenizerAt(i)->GetTokenAtAsString(4) == "p") ||
+                          (Parser.GetTokenizerAt(i)->GetTokenAtAsString(4) == "l"));
+      
+      CM_ROEToLine[R] = i;
+    }
+  }
+
+  // Get the parameters and use ROOTs built in TF1 to invert the polynomials
```
zoglauer: We are not yet inverting - that got me confused for a moment

src/MSubModuleStripReadout.cxx
```diff
@@ -71,13 +78,99 @@ bool MSubModuleStripReadout::Initialize()
...
+    }
+    //TODO: @RobinAnthonyPetersen add all the other types of fits melinator can do
+    // So far only added these ones because these are the ones we use for the ecals
+  }
+
   return MSubModule::Initialize();
 }
 
 
 ////////////////////////////////////////////////////////////////////////////////
```
zoglauer, line 167: Can you check that you don't double this line? Same below

src/MSubModuleStripReadout.cxx
fhagemann: The inversion happens later with `GetX`. I can move this comment to later.

src/MSubModuleStripReadoutNoise.cxx
fhagemann: Looks like this was initially introduced for adding micro-phonics, random triggers and nearest neighbors.
https://github.com/cositools/nuclearizer/blob/3621805d3b9cc6932ad4309f23d82da1698a07b4/src/MModuleDEESMEX.cxx#L179-L186

I would say, we keep it for now and can still remove it later.

src/MSubModuleStripReadout.cxx
fhagemann: Yes, this should be handling the case for shorted strips.

src/MSubModuleStripReadoutNoise.cxx
fhagemann: The question would just be: would it make sense to add random noise etc. already at the keV level, and only THEN convert to ADC (?)
Currently we would add all of this to the values already converted to ADC.

fhagemann:
The latest commit should address all comments, except for how to deal with the `MSubModuleReadoutTriggerNoise`.
I also added Robin to the copyright notices.

src/MSubModuleStripReadoutNoise.cxx
zoglauer: Probably after we have converted it to ADC.

zoglauer:
You can merge it.

PR #120, opened 2026-03-25
FIX: Reading calibrated timing in `MModuleLoaderMeasurementsROA`

fhagemann:
The ROA loader is currently assigning the values for `Timing` to `m_TAC` instead of `m_Timing`.
This is fixed now.

### DIFF
--- a/src/MModuleLoaderMeasurementsROA.cxx
+++ b/src/MModuleLoaderMeasurementsROA.cxx
@@ -208,7 +208,7 @@ bool MModuleLoaderMeasurementsROA::ReadNextEvent(MReadOutAssembly* Event)
     SH->SetStripID(Strip->GetStripID());
     
     if (Timing != nullptr) {
-      SH->SetTAC(Timing->GetTiming());
+      SH->SetTiming(Timing->GetTiming());
     }
     if (TAC != nullptr) {
       SH->SetTAC(TAC->GetTAC());


fhagemann:
Also: what is `m_Timing` supposed to be? The calibrated TAC value in nanoseconds? Or the drift time since the creation of electron hole pairs (in nanoseconds)?

According to the comment in `MStripHit`, it should be a `double` denoting the  `//! Timing in ns`
https://github.com/cositools/nuclearizer/blob/07e64cddf3433806a07518390e48a24b7d0a5452/include/MStripHit.h#L189-L190

In the `ReadOutAssemblyTiming` in `megalib`, `m_Timing` is an `unsigned int` denoted as `//! The timing`:
https://github.com/zoglauer/megalib/blob/c5fa4af354fc36be4cd71511402a68f0d6d6952f/src/fretalon/base/inc/MReadOutDataTiming.h#L90-L91

Currently, this would write floating point numbers to the file, but expects integers when reading them, resulting in very funky behavior.

ckierans:
Good catch!

m_Timing should be the TAC calibrated into ns. m_TAC was a variable Sean introduced awhile back. We used to only have the Timing keyword and it was overloaded to represent the ADC value *and* the ns conversion. m_Timing should be a floating point whereas m_TAC should be an int.

fhagemann:
But changing the type of `m_Timing` this something that we need to tackle in `megalib` (on the `develop-cosi` branch) 
then? If so, I can open an issue / draft a PR for that.

zoglauer:
The MEGalib part is what we read from the roa file, the TAC value.
Thus is should be an integer

fhagemann:
> The MEGalib part is what we read from the roa file, the TAC value. Thus is should be an integer

But let's assume we write calibrated timing values to ROA, what would be the preferred reader then?
I would then need to use `ReadOutDataTiming` for the `m_TAC` and something else (that would then still need to be written?) for `m_Timing`.
I feel the naming is a bit inconsistent here..

zoglauer:
Yes, Timing is from the old days. We didn't call it TAC in the balloon days. But why do we write ns timing to roa files? roa files are for the data from the raw measurements.

fhagemann:
Well, the current ROA loader has to possibility to read "TAC" and "Timing" and attributes it to the `m_TAC` and `m_Timing` value of the strip hits (?)

No my question would be if this way of reading the data is actually correct or not?
https://github.com/cositools/nuclearizer/blob/fc79014156600eaa81cee8b118b682f1e7c2249d/src/MModuleLoaderMeasurementsROA.cxx#L188-L233

fhagemann:
And there is also possibility to read/write `ADC` and `Energy`, the equivalent for the slow shaper output.

fhagemann:
This is a screenshot of what I have:
`adc` = (Uncalibrated) energies in ADC units
`tac` = (Uncalibrated) TAC values in ADC units
`energy` = (Calibrated) energies in keV
`timing` = (Calibrated) TAC values in ns

**Note that it's the same quantities, just differing in calibration/units. 
Timing right now would NOT be drift time or something.
It's still the TAC value, just in nanoseconds.**

image

PR #121, opened 2026-04-06
Saving Strip Pairing Chi Square for Each Detector

julianmgerber:
Summary of Changes:
- Storing information on best chi square for each input detector
- Printing out chi square for each detector in .dat file
- Using minimum/maximum chi square in vector in Event Filter module (as opposed to average of the chi square)

### DIFF
--- a/src/MReadOutAssembly.cxx
+++ b/src/MReadOutAssembly.cxx
@@ -541,12 +541,6 @@ bool MReadOutAssembly::StreamDat(ostream& S, int Version)
   S<<"ID "<<m_ID<<endl;
   S<<"CL "<<m_Time<<endl;
   S<<"TI "<<m_EventTimeUTC<<endl;
-  S<<"QP";
-  // iterate through the vectorized strip pairing reduced chi squares
-  for (auto i : m_StripPairingReducedChiSquare) {
-    S<<" "<<i;
-  }
-  S<<endl;
     
   for (MSimIA& IA: m_SimIAs) {
     S<<IA.ToSimString()<<endl; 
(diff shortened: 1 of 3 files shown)


src/MReadOutAssembly.cxx
```diff
@@ -531,12 +541,21 @@ bool MReadOutAssembly::StreamDat(ostream& S, int Version)
   S<<"ID "<<m_ID<<endl;
   S<<"CL "<<m_Time<<endl;
   S<<"TI "<<m_EventTimeUTC<<endl;
-  S<<"QP "<<m_StripPairingReducedChiSquare<<endl; // Read out strip pairing qualiy factor
+  S<<"QP";
+  // iterate through the vectorized strip pairing reduced chi squares
+  for (auto i : m_StripPairingReducedChiSquare) {
+    S<<" "<<i;
+  }
+  S<<endl;
```
ckierans: You only have the streaming defined for the .dat file, but we also will want this in the evta file - that's the standard format for revan. This should probably be incorporated in the StreamBDFlags function where you already have the Strip Pairing quality flag printout. Unless there's a reason you had this pulled out separately?

ckierans (commented):
Looks good, Julian. Just make sure this vector is being saved to the EVTA file so it can be filtered in mimrec.

src/MReadOutAssembly.cxx
```diff
@@ -531,12 +541,21 @@ bool MReadOutAssembly::StreamDat(ostream& S, int Version)
   S<<"ID "<<m_ID<<endl;
   S<<"CL "<<m_Time<<endl;
   S<<"TI "<<m_EventTimeUTC<<endl;
-  S<<"QP "<<m_StripPairingReducedChiSquare<<endl; // Read out strip pairing qualiy factor
+  S<<"QP";
```
zoglauer: Can you switch this to PQ? 
It turns out we did this already for NCT test wise, and this flag already exists in MEGAlib, I just have to modify it too accept multiple numbers.

ckierans:
Looks good, Julian!

PR #123, opened 2026-04-14
FIX: Update ASIC polarities only for HDF5 version 2

fhagemann:
I am fixing a stupid mistake I introduced.
Now: the polarities of the ASICs are only updated based on the HDF5 files if they are minimum version 2


ckierans:
Thanks for the quick fix!

PR #124, opened 2026-04-14
Timing cleanup

ckierans:
I scrubbed the modules to remove balloon timing variables CL and TI, and introduced a new variable RTS for Reference Time Standard (mission time in seconds from Jan 1, 2025 in TT). We should now only have RTS time and UTC time for each event.

The RTS variable will only be used for the FITS formatted files (it will be passed to Nuclearizer from the L1a files). UTC time will be recorded as TI for each event in the EVTA/ROA/DAT files.

The calculation of EventTimeUTC from TimeCode for each HDF file version needs to be corrected in MModuleLoaderMeasurementsHDF.cxx once the time variable in the data is meaningful. I wasn't able to check this with older hdf file formats.

### DIFF
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -813,14 +813,14 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
 
     Event->SetID(LongEventID);
     if (m_HDFStripHitVersion == MHDFStripHitVersion::V1_0) {
-      Event->SetCL(TimeCode);
+      Event->SetTimeUTC(TimeCode);
     } else if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_0)  {
-      Event->SetTI(TimeCode);
+      Event->SetTimeUTC(TimeCode);
       if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_2) {
-        Event->SetCL(SPWTimeCode);
+        Event->SetTimeUTC(SPWTimeCode);
       }
     } else {
-      Event->SetTI(TimeCode);
+      Event->SetTimeUTC(TimeCode);
     }
   }
 
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -813,6 +813,7 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
     unsigned long LongEventID = EventID + m_NumberOfEventIDRollOvers*(numeric_limits<uint16_t>::max() + 1);
 
     Event->SetID(LongEventID);
+
     if (m_HDFStripHitVersion == MHDFStripHitVersion::V1_0) {
       TimeUTC.Set(TimeCode,0);
       Event->SetTimeUTC(TimeUTC);
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -611,6 +611,7 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
     uint16_t EventID;
     uint64_t TimeCode;
     uint8_t NumberOfHits;
+    MTime TimeUTC;
 
     // Setting SPWTimeCode default to 0, as it is defined only iin HDF version >= 2.2
     uint64_t SPWTimeCode = 0;
@@ -813,14 +814,17 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
 
     Event->SetID(LongEventID);
     if (m_HDFStripHitVersion == MHDFStripHitVersion::V1_0) {
-      Event->SetTimeUTC(TimeCode);
+      TimeUTC.Set(TimeCode,0);
+      Event->SetTimeUTC(TimeUTC);
     } else if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_0)  {
-      Event->SetTimeUTC(TimeCode);
-      if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_2) {
-        Event->SetTimeUTC(SPWTimeCode);
-      }
+      TimeUTC.Set(TimeCode,0);
+      Event->SetTimeUTC(TimeUTC);
+    } else if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_2) {
+      TimeUTC.Set(TimeCode,0);
+      Event->SetTimeUTC(TimeUTC);
     } else {
-      Event->SetTimeUTC(TimeCode);
+      TimeUTC.Set(TimeCode,0);
+      Event->SetTimeUTC(TimeUTC);
     }
   }
 
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -101,8 +101,12 @@ bool MModuleLoaderMeasurementsHDF::Initialize()
   m_FileType = "Unknown";
   m_Detector = "Unknown";
   m_Version = -1;
-  /*
+  
+  // Start time of the file taken from the file name
+  // to be used to find absolute time for Spacewire brick
+  //TODO: Get more accurate start time from data files?
   m_StartObservationTime = MTime(0);
+  /*
   m_EndObservationTime = MTime(0);
   m_StartClock = numeric_limits<long>::max();
   m_EndClock = numeric_limits<long>::max();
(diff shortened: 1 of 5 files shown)


ckierans:
Confirmed TI in evta file is now correct with HDF v1.2 and HDF v2.2 file types.

For **HDF v1.2**, the GSE_TIMECODE is in Unix time of when the packet was received at the GSE (i.e. not a great measure of the event time), and this is defined as the Event time with the UTCTime parameter.

For example, the first hit in gse_20250709T113820.hdf5 (HP52358-3, Cs137 run) has a GSE_TIMECODE of 1752086301.09983.

Now, with these changes, I see the first event in the evta file saved as:
> SE
> ID 1
> **TI 1752086301.099829912**
> HT 3;0.291016;0.330591;3.19117;92.3539;0.0336036;0.0270836;0.0336036;1.26495
> CC NStripHits 3

For **HDF v2.2**, the SPW_TIMECODE is the accurate event time in ns relative to the start of acquisition. Therefore, we need to also add the start time which we get from the file name.

Looking at gse_20260217T115220.hdf5 (HP52406-1 livetime tests), the second Event (the first doesn't make it into the evta file) has a SPW_TIMECODE of 890324798 (in ns), and the date in the filename translates to a UTC time of 1771347140.

With these changes, I see the first event in the evta file saved as:
> SE
> ID 2
> **TI 1771347140.890324798**
> HT 3;1.63427e+307;1.63427e+307;1.63427e+307;39.7932;1.63427e+307;1.63427e+307;1.63427e+307;1.31098
> HT 3;1.63427e+307;1.63427e+307;1.63427e+307;44.0267;1.63427e+307;1.63427e+307;1.63427e+307;1.25952
> CC NStripHits 4
> BD DepthCalibrationError (Multiple hits on single strip) (Multiple hits on single strip)
> QA StripPairing (Best strip pairing leaves at least one grouping of strips unpaired (unpaired energy: 55.715945)) (Event contains multiple hits on a single strip)

(Ignore the warnings since I loaded this detector file but I still have all of the calibration data from HP52358-3 in the processing pipeline)

Still TODO is check the RTS with FITS format and create converters to go from UTC<->RTS.

ckierans:
I've now confirmed I get the right timing values in the FITS L1b file when I read in with HDF and convert from UTC.

Still TODO is to test with a FITS L1a format which includes RTS timing.

src/MReadOutAssembly.cxx
```diff
@@ -414,18 +411,43 @@ void MReadOutAssembly::RemoveHit(unsigned int i)
 ////////////////////////////////////////////////////////////////////////////////
 
 
+MTime MReadOutAssembly::ComputeRTSfromUTCTime(MTime UTCTime)
+{
+  //! Compute the RTS time if the event only has UTC time defined
+  //! RTS is elapsed time since Jan 1, 2025 in TT
+  //! TT = UTC + 37 + 32.184
```
fhagemann, line 424: I just looked this up ^^
Do we expect any leap seconds to be added here in the near future? Or are we fair to assume that these `37` are well enough for the COSI SMEX time frame?

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -171,6 +175,23 @@ bool MModuleLoaderMeasurementsHDF::OpenHDF5File(MString FileName)
...
+      unsigned int Day = FileDateTime.GetSubString(6,2).ToInt();
+      unsigned int Hour = FileDateTime.GetSubString(9,2).ToInt();
+      unsigned int Min = FileDateTime.GetSubString(11,2).ToInt();
+      unsigned int Sec = FileDateTime.GetSubString(13,2).ToInt();
+      m_StartObservationTime = MTime(Year, Month, Day, Hour, Min, Sec, 0);
+      if (g_Verbosity >= c_Info) cout<<m_XmlTag<<": Found start time from file name (UTC): "<<m_StartObservationTime<<endl;
+    } else {
+      if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Unable to determine start time from file name: "<<FileName<<endl;
+      return false;
+    }
```
fhagemann, line 193: Do we worry about timezones here? The current code might give the correct time stamp in PDT/PST, but not necessarily in UTC

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -798,16 +821,19 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
-        Event->SetCL(SPWTimeCode);
-      }
+
+    // Define event time based on the timecode within the HDF versions
+    if (m_HDFStripHitVersion <= MHDFStripHitVersion::V2_0) {
+      TimeUTC.Set(TimeCode); // Timecode in early versions is GSE computer time in s since Epoch
+      Event->SetTimeUTC(TimeUTC);
+    } else if (m_HDFStripHitVersion >= MHDFStripHitVersion::V2_2) {
+      MTime SPWTimeforEvent(m_StartObservationTime.GetAsSystemSeconds(),SPWTimeCode); // Spacewire Timecode is ns since start of aquisition
+      Event->SetTimeUTC(SPWTimeforEvent);
```
fhagemann, line 845: Timezones?

fhagemann (commented):
Thanks Carolyn for this work!
Just some timing conversion questions:
- do we expect the number of leap seconds to convert TT<->UTC to change?
- do we care about timezones to convert filename timestamps (usually in PDT/PST) to UTC?

src/MReadOutAssembly.cxx
ckierans: It looks like this is an active area of dispute! 
https://en.wikipedia.org/wiki/Leap_second

"Due to a recent acceleration in Earth's rotation, no leap seconds are expected in the near future, and the practice is set to be abolished or significantly modified by 2035."

So I think we'll need to check back on this if/when there is an annoucement of another leap second addition.

src/MModuleLoaderMeasurementsHDF.cxx
ckierans: Solid point about timezones. I've been assuming that everything in the file name for final formats will be using UTC, but you're right that we've been using PDT/PST in the filenames.

I see that we get the UTC of when the data was saved in the Header information (which is equivalent to how we're using the filename for now), so I can instead extract the start time from there to be sure of the timezone. But we should probably be saving the filenames in UTC vs PDT/PST...

fhagemann: How exactly do you plan to implement this?
One issue I see with using the timestamp when the data was saved would be having multiple running files (`gse_XXXX.hdf5`, `gse_XXXX_1.hdf5`, `gse_XXXX_2.hdf5`, ...), where the start of the data acquisition might be the same for all of them, but the time the data was saved might not be :thinking:

ckierans: Thinking about this more, I think I'm going to leave it as is for now. 

Discussing with Brent, it was agreed that all EM and FM testing going forward will use UTC instead of pacific time zones. So this won't be an issues for EM/FM calibration data.

Since the absolute time (UTC vs pacific time zone conversions) isn't crucial for the measurements we've made so far, and we have very limited measurements done with v2.2 using pacific time zones, I see the filename grab as sufficient until we have more v2.2 data to test with. It at least solves the issues with multiple files that you mention above.

fhagemann: Alright, makes sense! :)

src/MModuleLoaderMeasurementsHDF.cxx
ckierans: Are you okay if we ignore timezone since everything other than the Unit Level tests will use UTC? 

I could go into the file and compare the UTC and GSE_TIMECODE of the first event to see if they differ by an integer number of hours and correct the time from the file name to UTC to be sure, but I'm not sure it's worth the effort considering everything will soon be UTC...

fhagemann: Yes, that's fine for me.
I don't think it's worth the extra time, especially if we expect timestamps in UTC for EM/FM testing.

ckierans:
I checked this with a L1A FITS file from Wing (COSI_L0_obs_1980_006_apid0x0DD_v01_l1a.fits). The time stamps in his L1a file are negative, but the match checks out in the RST conversion even though the final numbers in the evta file are nonsense. It will be good to confirm these numbers again when we have a real FITS L1a data file.

fhagemann (approved):

src/MReadOutAssembly.cxx
```diff
@@ -420,18 +417,43 @@ void MReadOutAssembly::RemoveHit(unsigned int i)
...
+}
+
+
+////////////////////////////////////////////////////////////////////////////////
+
+
 bool MReadOutAssembly::Parse(MString& Line, int Version)
 {  
   // Handles SE, TI, RO, IA
   if (MReadOutSequence::Parse(Line) == true) return true;
```
fhagemann, line 455: I think a final question here would be:
currently we're reading `TI` entries and storing them to `m_Time` of the `MReadOutSequence` base class rather than `m_EventTimeUTC` of the `MReadOutAssembly`.

Reposting my question from Slack: I think we haven't clearly defined what `TI` in these ROA files should be -- according to the documentation it just says "Time in seconds" without defining the time frame or any reference:

Image

fhagemann: For the DEE, I would need to add this line to get some output in `TI` in the ROA files
(basically just to pass on what stored in the `TI` lines in the *.sim file):

image

ckierans: I answered in slack, but I think we want to keep the base class as general as possible and only impose the UTC requirement in our COSI-specific/nuclearizer pipeline. At least that's my feeling. Andreas should answer.

I'll merge this as is, but we can revisit if we need to change the base class.

zoglauer: Agreed

fhagemann:
Interestingly, when running the unit tests via GitHub actions, and comparing TRA files to the reference files,
it looks like the timing values are off by 28800 seconds (which is exactly 8 hours):
image

Let me look into if/how the local time of the machine running the code affects the timing in nuclearizer.

fhagemann:
I can reproduce this error locally when changing the time zone of my laptop to Europe.
Investigating this further..

src/MReadOutAssembly.cxx
fhagemann: --> #169

PR #128, opened 2026-04-14
ADD: Read JSON config also from HDF files in version 1 format

### LINKED ISSUE #125: Read ASIC polarities from data in HDF version 1
With PR #123, the ASIC polarities are read from the JSON config in the HDF5 file (verison 2)
In principle, this information should also be there in HDFv1, but this still needs to be implemented.

fhagemann:
Addressing #125:

The JSON config should be stored in the `Config` Dataset in HDF format version 1.
To read the polarities, we can use the same regex parser that I used for version 2.

### DIFF
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -150,6 +150,11 @@ struct MHDFStripHitVersionString {
   char string_col[256];
 };
 
+//! The config string
+struct MHDFConfigString {
+  char string_col[1048576];
+};
+
 ////////////////////////////////////////////////////////////////////////////////
 
 
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -150,9 +150,9 @@ struct MHDFStripHitVersionString {
   char string_col[256];
 };
 
-//! The config string
-struct MHDFConfigString {
-  char string_col[1048576];
+//! The JSON config string
+struct MHDFJSONConfigString {
+  char string_col[262144];
 };
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -206,7 +206,7 @@ bool MModuleLoaderMeasurementsHDF::OpenHDF5File(MString FileName)
 
         // Create compound type for reading the JSON config string
         CompType ConfigType(sizeof(MHDFJSONConfigString));
-        ConfigType.insertMember("string_col", HOFFSET(MHDFJSONConfigString, string_col), StringType);
+        ConfigType.insertMember("string_col", HOFFSET(MHDFJSONConfigString, string_col), ConfigStringType);
 
         MHDFJSONConfigString CS;
         ConfigDataset.read(&CS, ConfigType);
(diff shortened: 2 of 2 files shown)


include/MModuleLoaderMeasurementsHDF.h
```diff
@@ -150,6 +150,11 @@ struct MHDFStripHitVersionString {
   char string_col[256];
 };
 
+//! The config string
+struct MHDFConfigString {
+  char string_col[1048576];
```
fhagemann: I chose 1048576 here.
Usually, the JSON config strings are around 200,000 to 300,000 bytes.

Image

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -194,6 +195,16 @@ bool MModuleLoaderMeasurementsHDF::OpenHDF5File(MString FileName)
...
 
+      DataSet ConfigDataset = m_HDFFile.openDataSet("/Config");
+
+      CompType ConfigType(sizeof(MHDFConfigString));
+      StrType ConfigStringType(PredType::C_S1, 1048576);
+      ConfigType.insertMember("string_col", HOFFSET(MHDFConfigString, string_col), StringType);
+
+      MHDFConfigString CS;
+      ConfigDataset.read(&CS, ConfigType);
+      ConfigJSON = string(CS.string_col);
```
fhagemann: This might still be missing a warning/error in case that the file does not have a valid `Config`.
I've heard that this might be the case for some GSE versions, where the first file might have the Config, but not all consecutive file
(e.g. `gse_XXX.hdf5` has a valid `/Config`, but `gse_XXX_1.hdf5` does not)

src/MModuleLoaderMeasurementsHDF.cxx
fhagemann: This would then also need to be adjusted to "if there are ASIC polarities"

include/MModuleLoaderMeasurementsHDF.h
fhagemann: Never mind, I just found in the specification that the `itemsize` is 262144. Will change this!
image

src/MModuleLoaderMeasurementsHDF.cxx
fhagemann: --> Apparently, there should always be `/Config` if there is also `/HDFVersion`,
so if we want to catch this missing in the HDF5 file, this would need to be earlier in the code.

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -174,7 +173,13 @@ bool MModuleLoaderMeasurementsHDF::OpenHDF5File(MString FileName)
...
-    // ToDo: Check for version.
+    // JSON config string containing the information on the ASIC polarities
+    string ConfigJSON;
+
+    // ToDo: Check for version
+    // Version 1.0 and 1.1 did not have /HDFVersion,
+    // Some GSE versions (6.1.1 ?) did not write /HDFVersion to EVERY file of a 
+    // multi-file measurement but only to the first one --> How to deal with this?
     m_HDFStripHitVersion = MHDFStripHitVersion::V1_0;
     if (H5Lexists(m_HDFFile.getId(), "HDFVersion", H5P_DEFAULT) > 0) {
```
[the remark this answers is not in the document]
fhagemann: Dismissing this, because whenever /HDFVersion does not exist, /Config also does not exist. No need to check for this.

ckierans:
Looks good, Felix. It's cool to have Copilot do a review for you! I'll have to try that...
I have no other comments. It's ready to merge when you are.

PR #130, opened 2026-04-15
FIX: bug in `MModuleDepthCalibration::Initialize`

fhagemann:
In `MModuleDepthCalibration::Initialize`, after trying to read metrology files, we check if the filename is `true` instead of if the loading of the file was successful.


ckierans:
Good catch! Thank you!

PR #131, opened 2026-04-15
[DEE] Implement smearing simulated strip energies based on measured FWHM curves

### LINKED ISSUE #119: Apply energy smearing based on the FWHM fit function in the ecal files to the DEE energies
Currently, we just convert energies to ADC in the DEE using the inverse energy calibration.
We might also need to use the FWHM energy resolution fit to smear those energies.

fhagemann:
Another DEE pull request where @robin4730 deserves all the credit!
It adds the functionality to smear the simulated strip energies after charge transport based on the FWHM curves defined in the ecal file.

This PR implements:
- parsing the FWHM curves from the ecal files
- smearing all hit energies based on the FWHM evaluated at that energies

The user can choose whether the energies should be smeared via the nuclearizer GUI (`"Add noise to data"`):
image

### DIFF
--- a/src/MModuleDEESMEX.cxx
+++ b/src/MModuleDEESMEX.cxx
@@ -75,7 +75,7 @@ MModuleDEESMEX::MModuleDEESMEX() : MModule()
   m_HasOptionsGUI = true;
   
   // Default to adding noise to the sim data
-  m_ResolutionCalibration = true;
+  m_ApplyResolutionCalibration = true;
 }
 
 
@@ -95,6 +95,7 @@ bool MModuleDEESMEX::Initialize()
 {
   // Set the geometry to the SubModules using it
   m_ChargeTransport.SetGeometry(m_Geometry);
+  m_StripReadout.SetApplyResolutionCalibration(m_ApplyResolutionCalibration);
 
   // Initialize the module 
 
@@ -179,7 +180,7 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
   // Step (7): Handle GeD charge transport to grid and voxelation into strips
   m_ChargeTransport.Clear();
   m_ChargeTransport.AnalyzeEvent(Event);
-  
+
   // Step (8): Handle the strip readout: energy -> ADCs
   // Also includes user selected energy resolution with the FWHM values from the ecal 
   m_StripReadout.Clear();
@@ -239,8 +240,8 @@ void MModuleDEESMEX::Finalize()
   m_ShieldReadout.Finalize();
   m_ShieldTrigger.Finalize();
   m_ChargeTransport.Finalize();
-  m_StripReadoutNoise.Finalize();
   m_StripReadout.Finalize();
+  m_StripReadoutNoise.Finalize();
   m_StripTrigger.Finalize();
   m_DepthReadout.Finalize();
   m_Output.Finalize();
@@ -282,9 +283,9 @@ bool MModuleDEESMEX::ReadXmlConfiguration(MXmlNode* Node)
   m_Output.ReadXmlConfiguration(Node);
   
   // Add noise button
-  MXmlNode* ResolutionCalibrationNode = Node->GetNode("ResolutionCalibration");
+  MXmlNode* ResolutionCalibrationNode = Node->GetNode("ApplyResolutionCalibration");
   if (ResolutionCalibrationNode != nullptr) {
-    m_ResolutionCalibration  = ResolutionCalibrationNode->GetValueAsBoolean();
+    m_ApplyResolutionCalibration  = ResolutionCalibrationNode->GetValueAsBoolean();
   }
 
   return true;
@@ -312,7 +313,7 @@ MXmlNode* MModuleDEESMEX::CreateXmlConfiguration()
   m_Output.CreateXmlConfiguration(Node);
   
   // Add noise button
-  new MXmlNode(Node, "ResolutionCalibration", m_ResolutionCalibration);
+  new MXmlNode(Node, "ApplyResolutionCalibration", m_ApplyResolutionCalibration);
 
   return Node;
 }
--- a/src/MModuleDEESMEX.cxx
+++ b/src/MModuleDEESMEX.cxx
@@ -180,7 +180,7 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
   m_ChargeTransport.Clear();
   m_ChargeTransport.AnalyzeEvent(Event);
   
-  // Step (8)): Simulate micro-phonics random noise for triggered strips & next neighbors
+  // Step (8): Simulate micro-phonics random noise
   // Do this before energy -> ADCs because the FWHMs from the ecal are in keV
   if (m_AddNoise == true) {
     m_StripReadoutNoise.Clear();
--- a/src/MModuleDEESMEX.cxx
+++ b/src/MModuleDEESMEX.cxx
@@ -105,6 +105,7 @@ bool MModuleDEESMEX::Initialize()
   if (m_ShieldReadout.Initialize() == false) return false;
   if (m_ShieldTrigger.Initialize() == false) return false;
   if (m_ChargeTransport.Initialize() == false) return false;
+  if (m_StripReadoutNoise.Initialize() == false) return false;
   if (m_StripReadout.Initialize() == false) return false;
   if (m_StripTrigger.Initialize() == false) return false;
   if (m_DepthReadout.Initialize() == false) return false;
@@ -240,6 +241,7 @@ void MModuleDEESMEX::Finalize()
   m_ShieldReadout.Finalize();
   m_ShieldTrigger.Finalize();
   m_ChargeTransport.Finalize();
+  m_StripReadoutNoise.Finalize();
   m_StripReadout.Finalize();
   m_StripTrigger.Finalize();
   m_DepthReadout.Finalize();
--- a/src/MModuleDEESMEX.cxx
+++ b/src/MModuleDEESMEX.cxx
@@ -75,7 +75,7 @@ MModuleDEESMEX::MModuleDEESMEX() : MModule()
   m_HasOptionsGUI = true;
   
   // Default to adding noise to the sim data
-  m_AddNoise = true;
+  m_ResolutionCalibration = true;
 }
 
 
@@ -180,16 +180,16 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
   m_ChargeTransport.Clear();
   m_ChargeTransport.AnalyzeEvent(Event);
   
-  // Step (8): Simulate micro-phonics random noise
-  // Do this before energy -> ADCs because the FWHMs from the ecal are in keV
-  if (m_AddNoise == true) {
+  // Step (8): Handle the strip readout: energy -> ADCs
+  // Also includes user selected energy resolution with the FWHM values from the ecal 
+  m_StripReadout.Clear();
+  m_StripReadout.AnalyzeEvent(Event);
+  
+  // Step (9): Simulate micro-phonics random noise
+  if (m_ResolutionCalibration == true) {
     m_StripReadoutNoise.Clear();
     m_StripReadoutNoise.AnalyzeEvent(Event);
   }
-
-  // Step (9): Handle the strip readout: energy -> ADCs
-  m_StripReadout.Clear();
-  m_StripReadout.AnalyzeEvent(Event);
 
   // Step (10): Handles triggers and guard ring vetoes, pre-scalers, calculate dead-time, add nearest neighbor noise, calculate random coincidence time
   m_StripTrigger.Clear();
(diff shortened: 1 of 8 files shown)


src/MModuleDEESMEX.cxx
```diff
@@ -175,16 +178,18 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
...
-  // Step (8): Handle the strip readout: energy -> ADCs
+  // Step (9): Handle the strip readout: energy -> ADCs
   m_StripReadout.Clear();
   m_StripReadout.AnalyzeEvent(Event);
 
-
-  // Step (9)): Simulate micro-phonics random noise for triggered strips & next neighbors
-  m_StripReadoutNoise.Clear();
-  m_StripReadoutNoise.AnalyzeEvent(Event);
-
```
fhagemann: We decided swapping the order of adding noise and converting keV to ADC, because the ecal files provide FWHM **in keV** based on energies **in keV**, so it felt like the natural choice to FIRST add the noise, and THEN convert the energies to ADC.

src/MSubModuleStripReadout.cxx
```diff
@@ -48,6 +48,10 @@ ClassImp(MSubModuleStripReadout)
 MSubModuleStripReadout::MSubModuleStripReadout() : MSubModule()
 {
   // Construct an instance of MSubModuleStripReadout
+  
+  // Set all modules, which have to be done before this module
+  //TODO: @RobinAnthonyPetersen, need to make sure the strip noise gets added first! 
+  //AddPreceedingModuleType(MAssembly::c_StripReadoutNoise, true);
```
fhagemann: @zoglauer Does something like this exist for `MSubModules`?

fhagemann:
After actually initializing and finalizing `MSubModuleStripReadoutNoise` in `MModuleDEESMEX`, I can show some plots with `"Add noise to data"` set to `true` (blue) and `false` (orange):
image

fhagemann:
Actually, @robin4730 will move the FWHM smearing code to `MSubModuleStripReadout`, so that we can have the `MSubModuleStripReadoutNoise` empty for now, and maybe add microphonics, random strips triggering etc. added in the future.

src/MModuleDEESMEX.cxx
```diff
@@ -180,17 +180,17 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
...
+  // Step (8): Handle the strip readout: energy -> ADCs
+  // Also includes user selected energy resolution with the FWHM values from the ecal 
+  m_StripReadout.Clear();
+  m_StripReadout.AnalyzeEvent(Event);
+  
+  // Step (9): Simulate micro-phonics random noise
+  if (m_ResolutionCalibration == true) {
     m_StripReadoutNoise.Clear();
     m_StripReadoutNoise.AnalyzeEvent(Event);
   }
```
fhagemann: I don't think we need that `if`-statement here anymore..
`m_ResolutionCalibration` might now only be affecting what's happening in `m_StripReadout`

src/MSubModuleStripReadout.cxx
```diff
@@ -29,8 +29,11 @@
 // Standard libs:
 
 // ROOT libs:
+#include "TRandom.h"
+#include "TMath.h"
 
 // MEGAlib libs:
+#include "MSubModule.h"
```
fhagemann: This is not required, it already exists in `MSubModuleStripReadout.h`

src/MSubModuleStripReadout.cxx
```diff
@@ -90,73 +93,130 @@ bool MSubModuleStripReadout::Initialize()
...
-        Parser.GetTokenizerAt(i)->IsTokenAt(1, "dss") == true) {
-
-      MReadOutElementDoubleStrip R;
-      R.SetDetectorID(Parser.GetTokenizerAt(i)->GetTokenAtAsUnsignedInt(2));
-      R.SetStripID(Parser.GetTokenizerAt(i)->GetTokenAtAsUnsignedInt(3));
-      R.IsLowVoltageStrip((Parser.GetTokenizerAt(i)->GetTokenAtAsString(4) == "p") ||
-                          (Parser.GetTokenizerAt(i)->GetTokenAtAsString(4) == "l"));
+    MTokenizer* T = Parser.GetTokenizerAt(i);
+    
+    // IFF user wants it, get the fits to smear the energies based on the FWHM from the ecal
```
fhagemann: Typo: `IFF` --> `If`

src/MSubModuleStripReadout.cxx
```diff
@@ -90,73 +93,130 @@ bool MSubModuleStripReadout::Initialize()
...
-
-      m_Calibration[CM.first] = melinatorfit;
-    } else {
-      // TODO: Add all the other types of fits melinator can do
-      // So far, only added these ones because these are the ones we use for the ecals
-      if (g_Verbosity >= c_Error) {
-        cout<<m_Name<<": Unhandled CalibratorType: "<<CalibratorType<<endl<<"Please update this module."<<endl;
+    
+    
+    // Get enegry calirbation fits second
```
fhagemann: Typo: `calirbation` --> `calibration`

src/MSubModuleStripReadout.cxx
```diff
@@ -181,12 +241,35 @@ void MSubModuleStripReadout::Clear()
 bool MSubModuleStripReadout::AnalyzeEvent(MReadOutAssembly* Event)
 {
   // Main data analysis routine, which updates the event to a new level
+  
+  double FWHMtoSigma = 2.0 * TMath::Sqrt(2.0 * TMath::Log(2.0));
```
fhagemann: this might become a `constexpr` (?)

src/MSubModuleStripReadout.cxx
```diff
@@ -181,12 +241,35 @@ void MSubModuleStripReadout::Clear()
...
 
   // Get low-voltage and high-voltage hits
   for (auto* Hits : { &Event->GetDEEStripHitLVListReference(), &Event->GetDEEStripHitHVListReference() }) {
     
     for (MDEEStripHit& SH : *Hits) {
-    
+      
+      // IFF the user wants it applied, apply the FWHM Guassian energy resolution
+      
+      if (m_ApplyResolutionCalibration) {
```
fhagemann:
```
if (m_ApplyResolutionCalibration == true) {
```

src/MSubModuleStripReadout.cxx
```diff
@@ -90,73 +93,130 @@ bool MSubModuleStripReadout::Initialize()
...
-
-      MReadOutElementDoubleStrip R;
-      R.SetDetectorID(Parser.GetTokenizerAt(i)->GetTokenAtAsUnsignedInt(2));
-      R.SetStripID(Parser.GetTokenizerAt(i)->GetTokenAtAsUnsignedInt(3));
-      R.IsLowVoltageStrip((Parser.GetTokenizerAt(i)->GetTokenAtAsString(4) == "p") ||
-                          (Parser.GetTokenizerAt(i)->GetTokenAtAsString(4) == "l"));
+    MTokenizer* T = Parser.GetTokenizerAt(i);
+    
+    // IFF user wants it, get the fits to smear the energies based on the FWHM from the ecal
+    if (m_ApplyResolutionCalibration) {
```
fhagemann:
```
if (m_ApplyResolutionCalibration == true) {
```

src/MSubModuleStripReadout.cxx
```diff
@@ -181,12 +241,35 @@ void MSubModuleStripReadout::Clear()
...
   for (auto* Hits : { &Event->GetDEEStripHitLVListReference(), &Event->GetDEEStripHitHVListReference() }) {
     
     for (MDEEStripHit& SH : *Hits) {
-    
+      
+      // IFF the user wants it applied, apply the FWHM Guassian energy resolution
+      
+      if (m_ApplyResolutionCalibration) {
+        // Look up the FWHM fit for this strip
+        if (m_ResolutionCalibration.count(SH.m_ROE) > 0) {
```
fhagemann: Maybe also add an `else` here with a warning (or info?) that no FWHM curve was found. I added this when I was debugging the original PR.

src/MSubModuleStripReadoutNoise.cxx
```diff
@@ -29,12 +29,9 @@
 // Standard libs:
 
 // ROOT libs:
-#include "TRandom.h"
-#include "TMath.h"
 
 // MEGAlib libs:
 #include "MSubModule.h"
```
fhagemann: I would even argue that this could also be removed from here.

fhagemann (commented):
Thanks @robin4730!
Here some small comments :)

src/MSubModuleStripReadout.cxx
robin4730: What if I make this: static const double FWHMtoSigma = 2.0 * TMath::Sqrt(2.0 * TMath::Log(2.0));

src/MSubModuleStripReadoutNoise.cxx
robin4730: I think this is needed for: static const double FWHMtoSigma = 2.0 * TMath::Sqrt(2.0 * TMath::Log(2.0))?

src/MSubModuleStripReadout.cxx
fhagemann: This was not addressed?

src/MSubModuleStripReadoutNoise.cxx
fhagemann: I was talking about the `#include "MSubModule.h"`, not the `TMath` ;)
`MSubModule.h` is already included in the header file `MSubModuleStripReadoutNoise.h`

fhagemann:
We might want to squash and merge this, once it's approved ^^

fhagemann:
Sorry for all the force-pushes, I accidentally added some code for the DEEIntake that weren't supposed to be included here -- they are now removed again :)

src/MModuleDEESMEX.cxx
```diff
@@ -73,6 +73,9 @@ MModuleDEESMEX::MModuleDEESMEX() : MModule()
   AddSucceedingModuleType(MAssembly::c_NoRestriction);
   
   m_HasOptionsGUI = true;
+  
+  // Default to adding noise to the sim data
+  m_ApplyResolutionCalibration = true;
```
ckierans, line 78: This is okay for now, but once we're past more of the debugging stage, we should remove this option entirely. We should always have the smearing enabled.

src/MSubModuleStripReadout.cxx
```diff
@@ -91,72 +93,129 @@ bool MSubModuleStripReadout::Initialize()
...
+          
+          TF1* resFit = new TF1("res_poly3", "[0] + [1]*x + [2]*x^2 + [3]*x^3", 0., HistMaxkeV);
+          resFit->SetParameters(a0, a1, a2, a3);
+          m_ResolutionCalibration[R] = resFit;
+          
+        } else {
+          // TODO: @RobinAnthonyPetersen add all the other types of fits melinator can do
+          // So far, only added these ones because these are the ones we use for the ecals
+          if (g_Verbosity >= c_Warning) {
+            cout << m_Name << ": Unknown FWHM calibrator type: " << ResolutionCalibrationType << endl;
```
ckierans, line 153: If the ResolutiionCalibrationType is something other than the 3 above, I think you should return an Error and return false.

src/MSubModuleStripReadout.cxx
ckierans: this should probably be a const

src/MSubModuleStripReadout.cxx
```diff
@@ -181,12 +240,43 @@ void MSubModuleStripReadout::Clear()
...
   // Get low-voltage and high-voltage hits
   for (auto* Hits : { &Event->GetDEEStripHitLVListReference(), &Event->GetDEEStripHitHVListReference() }) {
     
     for (MDEEStripHit& SH : *Hits) {
-    
+      
+      // If the user wants it applied, apply the FWHM Guassian energy resolution
+      if (m_ApplyResolutionCalibration == true) {
+        // Look up the FWHM fit for this strip
+        if (m_ResolutionCalibration.count(SH.m_ROE) > 0) {
```
ckierans: .count(SH.m_ROE) should return 1 if the calibration if found, and 0 if not. Right? Maybe make that more explicit. The >0 seems too open.

ckierans (commented):
Looks good @robin4730 and @fhagemann. I just have a couple of small comments and then It's good to merge.

src/MModuleDEESMEX.cxx
fhagemann: --> I opened an issue to keep track of this: https://github.com/cositools/nuclearizer/issues/138

fhagemann:
I implemented the feedback and re-ran everything through the DEE to check that it still does the correct thing:
image

src/MSubModuleStripReadout.cxx
fhagemann: I declare this now as a constant in the header file.

ckierans:
Okay, looks good. Please merge when you're ready.

PR #74, opened 2025-11-25
Including Metrology Mask Rotation

ckierans:
I've added the ability to include the Yaw value from the mask metrology measurements to account for rotations of the strips on the LV and HV side. 

If the MaskMetrology is enabled in the Depth Calibration module, then the position of the hit is determined by the intercept of the dominate (i.e. largest energy deposit) LV and HV strips, where the centroid of the strip and angle are read from the metrology file. The intercept is calculated in a new function GetStripIntersection, and the metrology file is loaded with a new function LoadMaskMetrologyFile.

TODO: Aldo's file has the HV strip ID flipped (as evident in the x,y position reported in the metrology files), so there is a  `int HVStripID_flipped = 63 - HVStripID;` placeholder until the files are fixed.

TODO: The initial X/Y position should be more rigorous to include charge sharing down the line. This method will need to be updated at that time.

I confirmed that the implementation in nuclearizer gives ARM values that are the same as without the correction (unfortunately, not actually measurably better).

PR #134, opened 2026-04-24
[DEE] Implement inverse depth calibration

fhagemann:
This is currently still WIP, with the aim of implementing a routine to calculate TAC values from depth splines and depth calibration coefficient files.

## Basic idea
in the depth calibration, we generate a CTD-to-depth relation from simulations (detector simulations + electronics simulations), and determine stretch and offset corrections by matching it to measured CTD distributions.

From simulations, we can also get the individual electron and hole timings (time between creating the electron and hole pairs and the fast shaper signal reaching its maximum):
image

We can then apply stretch and offset to those charge carrier drift times 
(here for example for a stretch of 1.08 and an offset of 10ns)
image

And then evaluate those corrected charge carrier drift times at the simulated depth in the detector, to then determine the TAC value in nanoseconds, and subsequently convert it to ADC units using the inverse TAC calibration.

## ToDo list
Here is a checklist to keep track of what is done and what still needs to be done:
- [x] Parse CTD-to-depth splines for a single detector
- [x] Parse CTD-to-depth splines for multiple detectors
- [x] Parse depth coefficient file for a single detector
- [x] Parse depth coefficient file for multiple detectors
- [ ] Ensure that the simulated charge drift time splines span over the whole detector depth
- [x] Apply stretch and offset to the charge drift time splines
- [x] Apply charge drift time spline to extract charge drift times from the simulated depth
- [x] Convert the charge drift times to TAC values in ns (dummy implementation with magic number 4500ns)
- [ ] Convert the charge drift times to TAC values in ns (more realistic implementation)
- [x] Inverse TAC cal to convert `SH.m_Timing` (TAC in ns) to `SH.m_TAC` (TAC in ADC units)


fhagemann:
I'm already getting a somewhat realistic CTD distribution from the simulations of the Am241 measurement in the Q0Q1 lollipop position of the EM cryostat:
image

This is still missing stretch and offset, as well as the possibility to simulate multiple detectors though :sweat_smile:

fhagemann:
Still need to add a check to ensure that the spline covers the whole detector depth (maybe not needed because we check for monotonous behavior of the CTD-to-depth-spline).
And maybe a more realistic conversion of drift times to TAC.

Also, this implementation assumes all strips to behave like main strips. Timing values of nearest neighbors might not be too accurate right now.

fhagemann:
This is what it looks like now after applying stretch and offset (and also TAC calibration):
image
using detector calibration files for 434-1 and the current EM massmodel (Alyson's version) for cosima.

This tiny secondary peak at positive CTD values is probably due to some pixels having weird stretch and offset values.

(Actually it looks like this, because I apply pixels with no depth splines, depth calibration coefficients or TAC calibration -- in particular the guard ring -- a standard TAC value of 0):
image
--> I'm just confused though why these TAC values of 0 would not be cut by the TAC cuts in the forward pipeline, so this is something I still want to investigate ^^

fhagemann:
(Last force push was just a rebase onto the current `develop/em`)

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
     }
   }
   list<MDEEStripHit>& HVHits = Event->GetDEEStripHitHVListReference();
   for (MDEEStripHit& SH: HVHits) {
-    if (SH.m_IsGuardRing == false) {
-      SH.m_TAC = 8000 + 2000*(1.0-SH.m_SimulatedRelativeDepth);
-      if (SH.m_TAC > 16383) SH.m_TAC = 16383;
+    int DetID = SH.m_ROE.GetDetectorID();
+    double Z = SH.m_SimulatedPositionInDetector.Z();
+    if (m_DepthGrid.count(DetID) > 0) {
```
ckierans: Above you have "if (m_DepthGrid.count(DetID) == 1) {" for the LV strips. Is there a reason this is different?

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+
+
+////////////////////////////////////////////////////////////////////////////////
+
+
+bool MSubModuleDepthReadout::LoadSplinesFile()
+{
+  // Input spline files should have the following format:
+  // ### DetID, HV, Temperature, Photopeak Energy
+  // depth, ctd, electron_drift_time, hole_drift_time
```
ckierans, line 310: With this new Spline file format, I assume we'll be using this same file for the forward pipeline too? Can you update the MModuleDepthCalibration::LoadSplineFile function to match this new format?

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+  // Pixel code (10000*det + 100*LVStrip + HVStrip), Stretch, Offset, Timing/CTD noise, Chi2 for the CTD fit (for diagnostics mainly)
+
+  MFile CoeffsFile;
+  if (CoeffsFile.Open(m_DepthCoefficientsFile) == false) {
+    cout << "ERROR in MSubModuleDepthReadout::LoadCoeffsFile: failed to open coefficients file." << endl;
+    return false;
+  }
+
+  MString Line;
+  while (CoeffsFile.ReadLine(Line)) {
```
ckierans: I know this isn't your code originally, but this should be `CoeffsFile.ReadLine(Line) == true`

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+  }
+
+  return true;
+}
+
+
+////////////////////////////////////////////////////////////////////////////////
+
+
+bool MSubModuleDepthReadout::LoadSplinesFile()
```
ckierans: It would be nice to be able to call this function and the next from MModuleDepthCalibration (once we get them to match, see my comment below) instead of duplicating this code. I'm no c++ expert, but some googling reveals that you can use a function from another class with a "composition." Maybe something to look into for some cleaner code?

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
   // Dummy code
```
ckierans: You can probably remove this line now :)

ckierans (commented):
This looks like a great start, Felix! 

My only "major" comment is that it'd be nice to call the file loading functions from the forward pipeline Modules so as to not duplicate code. 

Also, I think we need to think more about how to handle GR SHs. Maybe you have a plan here. The GR SHs don't have a defined TAC value in the data (obviously). Above you said that you're giving them TAC = 0, and that anything with TAC = 0 should be filtered out in the TACCut Module? Not sure if we want to leave the TAC value for these SH undefined, but either way, we shouldn't have them filtered out. 

I'm okay if you want to merge this to continue making progress, or I'd be happy to review again after your next updates.

include/MSubModuleDepthReadout.h
```diff
@@ -77,10 +107,33 @@ class MSubModuleDepthReadout : public MSubModule
   // protected members:
  protected:
 
+  //! Filename of the depth coefficients (stretch, offset, timing noise, ...)
```
zoglauer: Every member should have its description in a //! line

include/MSubModuleDepthReadout.h
```diff
@@ -77,10 +107,33 @@ class MSubModuleDepthReadout : public MSubModule
...
+  MString m_DepthCoefficientsFile;
```
zoglauer: It a FileName, not a File

include/MSubModuleDepthReadout.h
```diff
@@ -77,10 +107,33 @@ class MSubModuleDepthReadout : public MSubModule
...
+  unordered_map<int, vector<double>> m_Coeffs;
+  double m_Coeffs_Energy;
+
+  //! Filename of CTD->Depth splines
+  MString m_DepthSplinesFile;
```
zoglauer: Same as above it is a file name, not a file

src/MSubModuleChargeTransport.cxx
```diff
@@ -247,7 +248,14 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
 
   // Calculate strip ID by rounding down intentionally to avoid truncation towards zero
   // TODO: Include mask metrology information when calculating the strip ID from the position.
-  int ID = static_cast<int>(std::floor((P + PWidth/2.0) / Pitch));
+  int ID = static_cast<int>(std::floor((P + PWidth/2.0) / PPitch));
+
+  // Calculate the strip ID for the opposite side of the detector (and explicitly check for guard ring)
+  int OppositeStripID = static_cast<int>(std::floor((Q + QWidth/2.0) / QPitch));
+  if (std::abs(P) > PWidth/2.0 && std::abs(Q) > QWidth/2.0 && std::hypot(P, Q) > Radius) {
```
zoglauer, line 275: Can you explain this? Do we not have a guard ring hit if either X / P is outside the strip area OR Y/ Q is outside the strip are OR the radius condition is working?

Side note: Do we really have a radial condition here or a strip-based stair function, which approximates a circle?

src/MSubModuleChargeTransport.cxx
```diff
@@ -247,7 +248,14 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
 
   // Calculate strip ID by rounding down intentionally to avoid truncation towards zero
   // TODO: Include mask metrology information when calculating the strip ID from the position.
-  int ID = static_cast<int>(std::floor((P + PWidth/2.0) / Pitch));
+  int ID = static_cast<int>(std::floor((P + PWidth/2.0) / PPitch));
```
zoglauer, line 271: Do we not have to do any guards for ending up in the guard ring, here?

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+            }
+            SH.m_TAC = 0;
+          }
+          if (SH.m_TAC > 16383) {
+            SH.m_TAC = 16383;
+          }
+        } else {
+          if (g_Verbosity >= c_Info) {
+            cout<<"MSubModuleDepthReadout::AnalyzeEvent: No depth coefficient found for pixel with code "<<PixelCode<<endl;
+            SH.m_TAC = 0;
```
zoglauer: Shouldn't this be outside of the if()?

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+            SH.m_HasTriggered = false;
+          }
+        }
+      }
+    } else {
+      cout << "No Depth Spline for Event with DetID " << DetID << endl;
```
zoglauer: Can you check that ALL couts are guarded by if (g_Verbosity >= c_Info) { or warning or error

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+            SH.m_Timing = gRandom->Gaus(SH.m_Timing, CTD_Sigma / TMath::Sqrt(2.0));
+          }
+
+          // Apply the inverse TAC cal to obtain TAC in ADC units
+          int StripCode = SH.m_ROE.GetDetectorID() + SH.m_ROE.GetStripID();
+          if (m_TACCal.count(StripCode) == 1) {
+            vector<double> TACCal = m_TACCal[StripCode];
+            double TACCalSlope = TACCal[0];
+            double TACCalOffset = TACCal[1];
+            SH.m_TAC = (SH.m_Timing - TACCalOffset) / TACCalSlope;
```
zoglauer, line 274: m_TAC is unsigned int, thus make sure that SH.m_Timing >= TACCalOffset

src/MSubModuleDepthReadout.cxx
zoglauer: This is valid for all shared code. You can do in Initialize():

MModuleTACcut TACCut;
TACCut.SetTACCalFileName(m_TACCalFileName);
if (TACCut.LoadTACCalFile(m_TACCalFileName) == true){
// copy the data over
}

modify the modules that all the functions you need are public.

fhagemann:
Thanks @ckierans and @zoglauer for the feedback, I’ll look into implementing this tomorrow!

src/MSubModuleDepthReadout.cxx
```diff
@@ -94,17 +111,298 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+          SH.m_Timing = 4200.0 - HoleDriftTime;
+
+          // Smear the timing value based on the given CTD resolution
+          // --> divide by √2 to obtain TAC resolution from CTD resolution
+          if (m_ApplyTimingResolutionCalibration == true) {
+            SH.m_Timing = gRandom->Gaus(SH.m_Timing, CTD_Sigma / TMath::Sqrt(2.0));
+          }
+
+          // Apply the inverse TAC cal to obtain TAC in ADC units
+          int StripCode = SH.m_ROE.GetDetectorID() + 100 * SH.m_ROE.GetStripID();
```
zoglauer: Can you verify that this encoding is the same as in the TAC loader?

fhagemann: --> This is now using the same convention as the TAC loader with DetID -> side -> StripID instead of just the StripCode

src/MSubModuleDepthReadout.cxx
fhagemann: I have implemented this now for loading the depth calibration coefficients and the TAC calibration parameters. Updating the depth splines will require a bit more of work because I added two more columns (and I think, we still might want to allow for the current spline format for the forward pipeline, as I'm just adding two extra columns while keeping the first two columns the same).

src/MSubModuleDepthReadout.cxx
fhagemann: We can use the same file, but I would still allow for the current format, just because the additional columns (electron and hole drift times) are not required in the forward pipeline.
The forward pipeline can read multiple columns and interpret them as CTD splines for different pixels e.g.
I will look into how to update this.

fhagemann:
Thanks again @ckierans and @zoglauer for the feedback.
I have updated the code in the DEE to use the available parsers from the forward pipeline.

I'm currently comparing the DEE output after the latest changes to the output using the code before your reviews:
image

Seems like the version you were reviewing was actually faulty (the new output looks way more reasonable), and I'll check which commit / parser-update caused this change to verify that this is done correctly now.

src/MSubModuleDepthReadout.cxx
fhagemann: I will check if `(SH.m_Timing - TACCalOffset) / TACCalSlope >= 0` (to also cover the unlikely case that `TACCalSlope` is negative)

fhagemann:
Ok, looks like this change was introduced when replacing `LoadTACCalFile`..
image

I checked that all other refactors had no influence on this CTD distribution.
image

src/MSubModuleDepthReadout.cxx
```diff
@@ -214,9 +223,8 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
           }
 
           // Apply the inverse TAC cal to obtain TAC in ADC units
-          int StripCode = SH.m_ROE.GetDetectorID() + SH.m_ROE.GetStripID();
```
fhagemann: Ok, the old code was wrong because here I forgot to multiply the `DetID` with `10'000`.
The new implementation should be correct.

src/MSubModuleDepthReadout.cxx
fhagemann: With my latest commit 7172efc, I updated the CTD-spline parser in `MModuleDepthCalibration` to disallow understanding multiple columns as different CTD splines for different hit grades (which I discussed with Field was not really used so far anyhow).
This should now allow to pass files in both the old (2 columns) and DEE (4 columns) format to the forward pipeline, as only the first two columns will be used there.

src/MSubModuleChargeTransport.cxx
fhagemann: This done later in the code, right after defining all the physical constants:
https://github.com/cositools/nuclearizer/blob/f9c0cbf546113740f9891a30405e15004fe37be1/src/MSubModuleChargeTransport.cxx#L269-L271

Let me move the definitions of the physical constants elsewhere, so that the guards are not that separated.

src/MSubModuleChargeTransport.cxx
fhagemann: Right now, I'm reading all of this from the `m_Geometry`.
If we can get a strip-based stair function from the `m_Geometry` as well, I'll be happy to update this :)

src/MSubModuleChargeTransport.cxx
fhagemann: --> a0eb449

fhagemann:
Ok, this is again ready to be reviewed!
@ckierans @zoglauer

src/MSubModuleChargeTransport.cxx
fhagemann: Ok, I would propose to postpone this as part of including metrology data #74.

ckierans (commented):
The changes look good, Felix. Thanks for your work cleaning up the code! And it's great to see the CTD shape look more reasonable. I'm good with you merging this, but perhaps @zoglauer wanted another look.

src/MSubModuleDepthReadout.cxx
```diff
@@ -91,22 +153,213 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+          SH.m_Timing = 4200.0 - ElectronDriftTime;
+
+          // Smear the timing value based on the given CTD resolution
+          // --> divide by √2 to obtain TAC resolution from CTD resolution
+          if (m_ApplyTimingResolutionCalibration == true) {
+            SH.m_Timing = gRandom->Gaus(SH.m_Timing, CTD_Sigma / TMath::Sqrt(2.0));
+          }
+
+          // Apply the inverse TAC cal to obtain TAC in ADC units
+          if (m_TACCal.count(DetID) == 1 && m_TACCal[DetID].size() >= 2 && m_TACCal[DetID][1].size() >= StripID) {
```
zoglauer: Are you sure about this:
m_TACCal[DetID][1].size() >= StripID
equal Strip <= m_TACCal[DetID][1].size()
But shouldn't it be StripID < m_TACCal[DetID][1].size() - it's zero based

src/MSubModuleDepthReadout.cxx
```diff
@@ -91,22 +153,213 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+          SH.m_Timing = 4200.0 - HoleDriftTime;
+
+          // Smear the timing value based on the given CTD resolution
+          // --> divide by √2 to obtain TAC resolution from CTD resolution
+          if (m_ApplyTimingResolutionCalibration == true) {
+            SH.m_Timing = gRandom->Gaus(SH.m_Timing, CTD_Sigma / TMath::Sqrt(2.0));
+          }
+
+          // Apply the inverse TAC cal to obtain TAC in ADC units
+          if (m_TACCal.count(DetID) == 1 && m_TACCal[DetID].size() >= 1 && m_TACCal[DetID][0].size() >= StripID) {
```
zoglauer: Are you sure about this:
m_TACCal[DetID][0].size() >= StripID
equal Strip <= m_TACCal[DetID][0].size()
But shouldn't it be StripID < m_TACCal[DetID][0].size() - it's zero based

src/MSubModuleDepthReadout.cxx
zoglauer: Did you do that? I see it in line 195 but not here?

src/MSubModuleDepthReadout.cxx
```diff
@@ -91,22 +153,213 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+
+        // Apply stretch based on Eq. (3) in https://doi.org/10.1016/j.nima.2026.171332
+        // Apply no offset to the electron drift time --> add it fully to the hole signal
+        unsigned int StripID = SH.m_ROE.GetStripID();
+        int PixelCode = 10000*DetID + 100*SH.m_OppositeStripID + StripID;
+        if (m_Coeffs.count(PixelCode) == 1){
+
+          vector<double> Coeffs = m_Coeffs[PixelCode];
+          double Stretch = Coeffs[0];
+          double CTD_Sigma = Coeffs[2] * m_Coeffs_Energy / SH.m_Energy;
```
zoglauer: It's different from line 177 above?

zoglauer:
Looks good. After you check the 3 above things you can merge it.

src/MSubModuleDepthReadout.cxx
fhagemann: Yes, you're absolutely right.. This is my Julia-wired brain thinking differently about comparing array indices to array sized :sweat_smile:

src/MSubModuleDepthReadout.cxx
fhagemann: Yes, good catch! The coeffs file contains timing noise in units of sigma, but the reader in `MModuleDepthCalibration` converts them already to FWHM. Seems like I forgot to update that here when changing the code to use the file loader from the depth calibration

fhagemann: I also changed the code to divide the FWHM by `2.355` instead of the analytical `2*sqrt(2*log(2))` to exactly compensate for the multiplication by `2.355` in ` MModuleDepthCalibration`.

fhagemann:
Ok, looks like the depth coefficient file did not have a header, such that `m_Coeffs_Energy` remained at a default value of 0 and did not smear the timing values at all. This is what caused my previous plot to show no difference between old and new code.

> image

I now added a respective header and I see a difference now, which implies that the changes did something and that the new code should be doing the correct thing (the old code overestimated sigma by a factor of `2.355`, hence the wider peaks):
image

I will add some example "dummy" files to the `resources` folder for people to run the DEE, and will proceed with merging this PR afterwards.

PR #135, opened 2026-04-27
Strip pairing speed optimizations

zoglauer:
This speeds up strip pairing without changing the pairing logic. The total nuclearizer run was O(10) faster.

Main changes:
- remove exact duplicate strip combinations before evaluating them
- pass large vectors by reference instead of copying them
- update some combination handling in place instead of copying full nested containers

The biggest improvement came from deduplicating combinations. The next biggest improvement came from passing large data structures by reference.

I tested it on a 2.5M data set, the results are the identical.

Please test again.


julianmgerber (approved):
Tested it on ARM data, and it took about 11 minutes to run (~1 hour of data, ~10 million events)! Usually, it takes about 2 hours. I'm also seeing an identical output. 

Thank you for this, this helps a lot!

PR #136, opened 2026-04-27
BUG: Remove events in the HDF5 parser with no valid strip hits

### LINKED ISSUE #133: Nuclearizer Hangs when HDF Loader Include Nearest Neighbors is Checked
I installed a clean new versions for COSI tools & nuclearizer this way:

`cd ~/`
`/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/cositools/cosi-setup/main/setup.sh) --branch=develop-cosi"`
`source source.sh `
`cd nuclearizer/`
`git checkout develop/em`
`git pull`
`make clean`
`make -j 15`
`cd ~/COSItools/`
`git clone https://github.com/cositools/massmodel-cosi-sttc.git`
`nuclearizer -c ~/cosi/analysis/config/nuclearizer_int.cfg `

MEGAlib commit: 27d84af6bb983bb573082259392aacc86dc95686
Nuclearizer commit: edcbf9f4003b5254246a6828bdc76dcc872aa52a
massmodel-cosi-sttc commit: 911512da4494568cce7646a53184eda7cb50519c

Attaching nuclearizer config file and other related files.

fhagemann:
Addressing the bug in the HDF5 parser seen in #133:

All of the events with bogus event ID seemed to have no strip hits.
image

I think what is happening that these Events end up having no (valid) strip hits, and therefore never get assigned any event ID in the HDF5 parser, resulting in uninitialized (bogus) event IDs.

The most straight-forward fix would be to remove those `Events` with no valid strip hits.
After this fix, I do not see these events with bogus IDs (4294967284) anymore.

### DIFF
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -321,14 +321,17 @@ bool MModuleLoaderMeasurementsHDF::OpenHDF5File(MString FileName)
     if (PropertyList.getLayout() == H5D_CHUNKED) {
       hsize_t ChunkDims[H5S_MAX_RANK];
       PropertyList.getChunk(Rank, ChunkDims);
-
-      cout<<"Chunk dimensions: ";
-      for (int i = 0; i < Rank; ++i) {
-        cout<<ChunkDims[i]<<" ";
+      if (g_Verbosity > c_Info) {
+        cout<<"Chunk dimensions: ";
+        for (int i = 0; i < Rank; ++i) {
+          cout<<ChunkDims[i]<<" ";
+        }
+        cout<<endl;
       }
-      cout<<endl;
     } else {
-      cout<<"Dataset is not chunked (layout is not H5D_CHUNKED)."<<endl;
+      if (g_Verbosity > c_Info) {
+        cout<<"Dataset is not chunked (layout is not H5D_CHUNKED)."<<endl;
+      }
     }
 
     if (m_HDFStripHitVersion == MHDFStripHitVersion::V1_0) {
@@ -752,11 +755,12 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
       // Create objects for all hits that belong to that event
       for (uint32_t i = EventIndices.m_FEEHits[0]; i < EventIndices.m_FEEHits[1]; i++) {
 
-        uint32_t IndexInBatch = i - m_MinHitIndex;
         if (i < m_MinHitIndex || i >= (m_MinHitIndex + m_Buffer_2.size())) {
           if (g_Verbosity >= c_Error) cout << m_XmlTag << ": Entry " << i << " is NOT in the current FEEHits buffer!" << endl;
           return false;
         } 
+
+        uint32_t IndexInBatch = i - m_MinHitIndex;
         
         MHDFStripHit_V2& Hit = m_Buffer_2[IndexInBatch];
         if (m_StripMap.HasReadOutID(Hit.m_StripID) == true) {
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -711,6 +711,17 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
         if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Read-out ID "<<StripID<<" not found in strip map"<<endl;
         return false;
       }
+
+      // Remove incomplete events (fewer strip hits than what is listed in HITS)
+      if (StripHitIndex > 0 && NumberOfHits != NStripHits) {
+        if (g_Verbosity >= c_Error) {
+          cout<<m_XmlTag<<": Event "<<Event->GetID()<<" had fewer strip hits ("<<StripHitIndex<<") than expected ("<<NStripHits<<"). Ignoring event."<<endl;
+        }
+        // Reduce the batch index and current hit counter to still process the hit from the next event
+        m_CurrentBatchIndex--;
+        m_CurrentHit--;
+        return false;
+      }
       
     } else if (m_HDFStripHitVersion <= MHDFStripHitVersion::V2_2) {
 
@@ -779,17 +790,6 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
       }
     } else {
       if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Unhandled HDF hit version found: "<<m_HDFStripHitVersion<<endl<<"Please update this module."<<endl;
-      return false;
-    }
-
-    // Remove incomplete events (fewer strip hits than what is listed in HITS)
-    if (StripHitIndex > 0 && NumberOfHits != NStripHits) {
-      if (g_Verbosity >= c_Error) {
-        cout<<m_XmlTag<<": Event "<<Event->GetID()<<" had fewer strip hits ("<<StripHitIndex<<") than expected according to the HDF5 file ("<<NStripHits<<")."<<endl;
-      }
-      // Reduce the batch index and current hit counter to still process the hit from the next event
-      m_CurrentBatchIndex--;
-      m_CurrentHit--;
       return false;
     }
 
--- a/src/MModuleLoaderMeasurementsHDF.cxx
+++ b/src/MModuleLoaderMeasurementsHDF.cxx
@@ -725,6 +725,10 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
         m_CurrentHit--;
         return false;
       }
+
+      // Increase counters
+      NStripHits = static_cast<unsigned int>(NumberOfHits);
+      StripHitIndex++;
       
     } else if (m_HDFStripHitVersion <= MHDFStripHitVersion::V2_2) {
 
@@ -791,6 +795,9 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
           if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Read-out ID "<<Hit.m_StripID<<" not found in strip map"<<endl;
           return false;
         }
+
+        // Use StripIndex here (without updating NStripHits) to exit the while-loop after finalizing the Event
+        StripHitIndex++;
       }
     } else {
       if (g_Verbosity >= c_Error) cout<<m_XmlTag<<": Unhandled HDF hit version found: "<<m_HDFStripHitVersion<<endl<<"Please update this module."<<endl;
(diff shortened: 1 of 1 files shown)


zoglauer:
You probably shold create an error message too:

if (g_Verbosity >= c_Error) cout<<m_XmlTag<<":Bla bla"<<endl;

Did you also see events with duplicate event IDs? Because those caused the original hang...

fhagemann:
Let me check

fhagemann:
Ok, I still get duplicate event IDs, but it also looks like the hits are assigned to the wrong events.

This is from the raw dataframe (columns are EVENT_ID, ENERGY_DATA, TIMING_DATA):
image

And this is what ends up in the ROA file:
Screenshot from 2026-04-27 10-09-16

I will look into this.

fhagemann:
Ok, I figured out where the bug with duplicate event IDs comes from.

It looks like some events have fewer strip hits than what is stored in `HITS`, e.g. in `gse_20260217T125741.hdf5` the event with ID 23971:
Screenshot from 2026-04-27 13-58-35

In the current logic, it continues adding strip hits until the number of expected strip hits is reached,
and falsely including the first strip hit from the following event:
image

As we assign the `EventID` based on the last strip hit, this then false gets the event ID from the following event.

A tricky thing: when excluding nearest neighbors, we actually do NOT always update the event ID, as we continue to the next step in the `while`, preventing the update of event ID.

https://github.com/cositools/nuclearizer/blob/52b10a8dd87a75b46ef049ef024cad31a20095d3/src/MModuleLoaderMeasurementsHDF.cxx#L705-L711

image

This is what might have caused the duplicate event IDs.

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -704,13 +704,9 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
         // NOTE: at some point we will want to remove this code and always include nearest neighbor data
         if (m_IncludeNearestNeighbor == false && HitType == 1) {
           delete H; // Clean up the memory we just allocated
-          // Increase counters
-          NStripHits = static_cast<unsigned int>(NumberOfHits);
-          StripHitIndex++;
-          continue;
```
fhagemann, line 710: I'm removing this `continue` statement such that we end up doing the data handling at the end of each iteration in the big `while`-loop.

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -790,6 +782,16 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
...
 
+    if (StripHitIndex > 0 && NumberOfHits != NStripHits) {
+      if (g_Verbosity >= c_Error) {
+        cout<<m_XmlTag<<": Event "<<Event->GetID()<<" had fewer strip hits ("<<StripHitIndex<<") than expected according to the HDF5 file ("<<NStripHits<<")."<<endl;
+      }
+      // Reduce the batch index and current hit counter to still process the hit from the next event
+      m_CurrentBatchIndex--;
+      m_CurrentHit--;
+      return false;
+    }
```
fhagemann: Now, if `StripHitIndex > 0` (meaning that we already had hits in that given `Event`) and the `NumberOfHits` in the file does not match the `NStripHits` that we got from previous hits, we remove this event.

This produces quite some output (maybe this should become an `c_Info` rather than an `c_Error`?):

Image

From my quick glimpse at the raw dataframe, this seems to flag the incomplete events well, e.g.


Image
Image

zoglauer:
Do you have any clue if all HDF4 v2.x data sets are affected? Or just the one from Brent?

fhagemann:
The HDF5 v2.x files follow a very different logic internally, so this is hard to say.
I haven't seen any "zero-data bugs" there so far, but I can check if there are events that have fewer number of strip hits that what they claim to have.

zoglauer:
OK. Leave the error output on for the time being.
Are you done with the changes? Can I review it?

fhagemann:
Yes, this is ready for review! Thanks :)

zoglauer:
Claude and codex have some comments in general on the HDF5 loader (I did not review it - they are not always right):

  1. High: include/MModuleLoaderMeasurementsHDF.h:116 and include/MModuleLoaderMeasurementsHDF.h:128 declare m_GSETimeCode / m_SPWTimeCode as uint64_t, but the HDF5 compound members are declared as
     PredType::IEEE_F64LE in src/MModuleLoaderMeasurementsHDF.cxx:392, src/MModuleLoaderMeasurementsHDF.cxx:401, and src/MModuleLoaderMeasurementsHDF.cxx:402. That means the file’s floating-point time values
     are being read into integer fields with the wrong in-memory type layout, so V2 timing data is effectively corrupted before it is used.
  2. High: src/MModuleLoaderMeasurementsHDF.cxx:733 through src/MModuleLoaderMeasurementsHDF.cxx:738 never assign SPWTimeCode for V2.2, yet src/MModuleLoaderMeasurementsHDF.cxx:808 stores it into
     Event->SetCL(SPWTimeCode). So every V2.2 event currently gets CL = 0, regardless of the file contents.
  3. Medium: src/MModuleLoaderMeasurementsHDF.cxx:755 computes uint32_t IndexInBatch = i - m_MinHitIndex; before checking i < m_MinHitIndex at src/MModuleLoaderMeasurementsHDF.cxx:756. Since these are
     unsigned values, the subtraction can underflow first. The subsequent check prevents the bad index from being used in the normal path, so this is not as severe as the timing bugs, but the check should
     come first.
  4. Low: src/MModuleLoaderMeasurementsHDF.cxx:325 and src/MModuleLoaderMeasurementsHDF.cxx:331 print chunking diagnostics unconditionally. Everything else in this loader is mostly verbosity-gated, so this
     will spam stdout for every open even in normal use.

  Open question, not a confirmed bug:

  - In the V2 path, the code intentionally uses HitEvent.m_GSETimeCode for TimeCode at src/MModuleLoaderMeasurementsHDF.cxx:731 and src/MModuleLoaderMeasurementsHDF.cxx:736, then stores that into
    Event->SetTI(...) at src/MModuleLoaderMeasurementsHDF.cxx:807. That may be intended, but once the type bug is fixed it is worth confirming whether TI is supposed to receive gse_timecode or the event
    timecode field.

fhagemann:
>     1. High: include/MModuleLoaderMeasurementsHDF.h:116 and include/MModuleLoaderMeasurementsHDF.h:128 declare m_GSETimeCode / m_SPWTimeCode as uint64_t, but the HDF5 compound members are declared as
>        PredType::IEEE_F64LE in src/MModuleLoaderMeasurementsHDF.cxx:392, src/MModuleLoaderMeasurementsHDF.cxx:401, and src/MModuleLoaderMeasurementsHDF.cxx:402. That means the file’s floating-point time values
>        are being read into integer fields with the wrong in-memory type layout, so V2 timing data is effectively corrupted before it is used.
--> Timing cleanup, tackled in #124 

>     2. High: src/MModuleLoaderMeasurementsHDF.cxx:733 through src/MModuleLoaderMeasurementsHDF.cxx:738 never assign SPWTimeCode for V2.2, yet src/MModuleLoaderMeasurementsHDF.cxx:808 stores it into
>        Event->SetCL(SPWTimeCode). So every V2.2 event currently gets CL = 0, regardless of the file contents.
--> also Timing cleanup, tackled in #124 

>     3. Medium: src/MModuleLoaderMeasurementsHDF.cxx:755 computes uint32_t IndexInBatch = i - m_MinHitIndex; before checking i < m_MinHitIndex at src/MModuleLoaderMeasurementsHDF.cxx:756. Since these are
>        unsigned values, the subtraction can underflow first. The subsequent check prevents the bad index from being used in the normal path, so this is not as severe as the timing bugs, but the check should
>        come first.
> 
>     4. Low: src/MModuleLoaderMeasurementsHDF.cxx:325 and src/MModuleLoaderMeasurementsHDF.cxx:331 print chunking diagnostics unconditionally. Everything else in this loader is mostly verbosity-gated, so this
>        will spam stdout for every open even in normal use.
--> both tackled in d9de69e
> 
> Open question, not a confirmed bug:
> 
>     * In the V2 path, the code intentionally uses HitEvent.m_GSETimeCode for TimeCode at src/MModuleLoaderMeasurementsHDF.cxx:731 and src/MModuleLoaderMeasurementsHDF.cxx:736, then stores that into
>       Event->SetTI(...) at src/MModuleLoaderMeasurementsHDF.cxx:807. That may be intended, but once the type bug is fixed it is worth confirming whether TI is supposed to receive gse_timecode or the event
>       timecode field.
--> also Timing cleanup, tackled in #124

zoglauer (approved):
Approved for merge

fhagemann:
> Do you have any clue if all HDF4 v2.x data sets are affected? Or just the one from Brent?

There also seem to be events with fewer hits than expected in HDFv2.3 (not many, here 4 out of almost 900,000):
```
gse_20260217T124025.hdf5 (HDFv2.3) with 895056 events:

Event 35546 had fewer strip hits (4) than expected (15).
Event 79597 had fewer strip hits (4) than expected (15).
Event 536325 had fewer strip hits (3) than expected (15).
Event 871071 had fewer strip hits (3) than expected (15).
```

fhagemann:
Let me have a quick look at HDF5 v2.x to see that everything works as intended before merging this ^^

fhagemann:
With the latest push, this should also work with HDFv2.

Because of the difference in internal structure of the HDF files in v1 and v2, the current implementation of reading strip hits in `AnalyzeEvent` is a bit dirty:
- For HDF5v1, we iterate through all entries of `Hits`, which are actual strip hits. We therefore require this `while`-loop to accumulate strips until the expected number of strip hits is reached.
- For HDF5v2, we iterate through all entries of `Events`/`EventIndices`, which should already include the indices for all strip hits (`FEEHits`). Therefore, we don't need to accumulate strip hits in the `while`-loop -- we just need to exit it at some point.

I added some comments to the code, but let me know if we should improve the code logic
(if so, then maybe let's still merge this PR and open a new issue for internal refactoring).

src/MModuleLoaderMeasurementsHDF.cxx
```diff
@@ -745,11 +759,12 @@ bool MModuleLoaderMeasurementsHDF::AnalyzeEvent(MReadOutAssembly* Event)
       // Create objects for all hits that belong to that event
       for (uint32_t i = EventIndices.m_FEEHits[0]; i < EventIndices.m_FEEHits[1]; i++) {
```
fhagemann, line 760: What's currently not in this PR, but what I could add:
checking that the number of `FEEHits` in the `EventIndices` matched the value of `hits` in `Events`. However, I am not sure how that might affect reading in data that has additional non-germanium hits (ACSHits, ...)

ckierans: I don't think we'll ever get FEEHits in the same EventID as ACSHits. The readout from the SCBs is separate, I think. We should confirm with Clio/Eric.

But this also makes me realize we won't have Event IDs in our final flight data formats. So we will need to eventually make sure our pipelines run okay with that omission. That's a later problem though.

I'm okay with merging as you have it, and then cleaning up the code logic at a later time (I'm sure this isn't the last time you're going to need to touch the file).

fhagemann:
Then, I'm also fine with merging this as is (and dealing with a possible refactoring later).

PR #142, opened 2026-05-05
CHG: Print DEE trigger diagnostics only if verbosity is set to `Info`

fhagemann:
Currently, when running the DEE, the strip and shield trigger module statistics are **always** printed streamed to `cout`.
With this PR, this will only be the case if the verbosity is explicitly set to `Info`.


ckierans:
Looks good. Thanks, @fhagemann !

PR #145, opened 2026-05-13
[DEE] Consistenly use `GetTimeUTC` instead of `GetTime`

fhagemann:
Following up on https://github.com/cositools/nuclearizer/pull/124#discussion_r3196618854, I looked up all occurences of `GetTime` in the DEE and replaced them with `GetTimeUTC`.

In order for this to work, the "biggest" change in this PR is adding the following line to `MSubModuleDEEIntake`:
```
Event->SetTimeUTC(Event->GetSimulatedEvent()->GetTime());
```
Before this PR, all `TI` values in the ROA files created using simulated DEE events were all zero (reason being: cosima `TI` values are read in and stored in `Event->GetSimulatedEvent->m_Time`, but we save `Event->m_EventTimeUTC` to the ROA file).

After this PR, when writing ROA files using simulated DEE events, the `TI` values in the resulting ROA files match the `TI` values in the input cosima file.

### DIFF
--- a/src/MModuleDEESMEX.cxx
+++ b/src/MModuleDEESMEX.cxx
@@ -130,7 +130,7 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
   m_Intake.Clear();
   m_Intake.AnalyzeEvent(Event);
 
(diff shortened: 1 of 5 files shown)


ckierans:
Looks good! Thanks, Felix!

PR #146, opened 2026-05-18
[DEE] Raise enable with hard ACS veto

### LINKED ISSUE #94: Raise enable after veto
Raise enable and clear any dead time flags after veto. 

Adding this as an issue so we can keep track of it.

parshadkp:
Few things to address in GR and shield veto to make it more realistic.

- [x] Any ACS hard veto should clear the dead time original dead time as the events will not be read out. Clear strip events in 1 µs dead time with a hard ACS veto (shield hit) as they will not be read out.

- [x] Remove m_ShieldVetoWindowSize and change to be m_ShieldVetoTime (event time of the veto) + m_DeadTimeEnd.

- [x] Add a toggle to control if GR and/or shield veto is active in GUI and config file.

- [x] m_ShieldVetoWindowSize should actually be the dead time of the ACS after a shield veto. m_ShieldVetoWindowSize will be converted to a number that is added/subtracted onto the dead time for any adjustments to the VetoWindowSize.


resource/calibration/COSI_SingleDet/deadtimeParameters.txt
```diff
@@ -1,4 +1,4 @@
-CoincidenceWindow (s); StripDeadtimePerStrip (s); delayAfter1 (s); delayAfter2 (s)
-1.4e-6		       1e-6	                  0.2e-6	   0.2e-6
-ShieldDeadtimePerStrip (s); ShieldVetoWindow (s); ShieldPulseDuration (s); ShieldDelayBefore (s); ShieldDelayAfter (s)
-1e-6                        1.5e-6                1.7e-6                   0.1e-6                 0.4e-6
+CoincidenceWindow (s); StripDeadtimePerStrip (s); delayAfter1 (s); delayAfter2 (s); FastClearDeadtime (s)
+1.4e-6		       1e-6	                  0.2e-6	   0.2e-6          1e-6
+ShieldDeadtimePerStrip (s); ShieldPulseDuration (s); ShieldDelayBefore (s); ShieldDelayAfter (s); ShieldVetoWindowDetla (s)
```
fhagemann:
```suggestion
ShieldDeadtimePerStrip (s); ShieldPulseDuration (s); ShieldDelayBefore (s); ShieldDelayAfter (s); ShieldVetoWindowDelta (s)
```

src/MSubModuleShieldTrigger.cxx
```diff
@@ -58,6 +58,8 @@ MSubModuleShieldTrigger::MSubModuleShieldTrigger() : MSubModule()
   m_HasTrigger = false;
   m_HasShieldVeto = false;
   m_IsShieldDead = false;
+  m_ShieldVetoTime = 0.0;
+  m_ShieldVetoWindowDetla = 0.0;
```
fhagemann:
```suggestion
  m_ShieldVetoWindowDelta = 0.0;
```

src/MSubModuleShieldTrigger.cxx
```diff
@@ -275,16 +278,16 @@ bool MSubModuleShieldTrigger::ParseDeadtimeFile()
...
 
   m_ASICDeadTimePerChannel = ShieldTokenizer->GetTokenAtAsDouble(0);
-  m_ShieldVetoWindowSize = ShieldTokenizer->GetTokenAtAsDouble(1);
-  m_ShieldPulseDuration = ShieldTokenizer->GetTokenAtAsDouble(2);
-  m_ShieldDelayBefore = ShieldTokenizer->GetTokenAtAsDouble(3);
-  m_ShieldDelayAfter = ShieldTokenizer->GetTokenAtAsDouble(4);
+  m_ShieldPulseDuration = ShieldTokenizer->GetTokenAtAsDouble(1);
+  m_ShieldDelayBefore = ShieldTokenizer->GetTokenAtAsDouble(2);
+  m_ShieldDelayAfter = ShieldTokenizer->GetTokenAtAsDouble(3);
+  m_ShieldVetoWindowDetla = ShieldTokenizer->GetTokenAtAsDouble(4);
```
fhagemann:
```suggestion
  m_ShieldVetoWindowDelta = ShieldTokenizer->GetTokenAtAsDouble(4);
```

include/MSubModuleShieldTrigger.h
```diff
@@ -145,8 +149,8 @@ class MSubModuleShieldTrigger : public MSubModule
   double m_ShieldDelayBefore;
   //! Shield delay 2 before trigger in seconds
   double m_ShieldDelayAfter;
-  //! Shield veto window size in seconds
-  double m_ShieldVetoWindowSize;
+  //! Adjustment to shield veto window in seconds
+  double m_ShieldVetoWindowDetla;
```
fhagemann:
```suggestion
  double m_ShieldVetoWindowDelta;
```

src/MSubModuleShieldTrigger.cxx
```diff
@@ -300,19 +303,34 @@ bool MSubModuleShieldTrigger::AnalyzeEvent(MReadOutAssembly* Event)
...
+      m_ShieldVetoTime = m_ShieldLastHitTime[group];
+      HasShieldVetoStart = true;
     }
   }
 
-  // Process shield hits and check for veto conditions
-  ProcessShieldHits(Event);
+  // Then veto events based on the max shield deadtime from that start time
+  if (HasShieldVetoStart == true &&
+      m_EventTime <= m_ShieldVetoTime + maxShieldDeadtime + m_ShieldVetoWindowDetla) {
```
fhagemann:
```suggestion
      m_EventTime <= m_ShieldVetoTime + maxShieldDeadtime + m_ShieldVetoWindowDelta) {
```

resource/calibration/COSI_SingleDet/deadtimeParameters.txt
fhagemann, line 1: I would have expected this file to be present in `resource/dee` if it's for DEE simulations.

src/MGUIOptionsDEESMEX.cxx
fhagemann, line 1: This looks nice and clean, I like it!

Image

src/MModuleDEESMEX.cxx
```diff
@@ -76,6 +76,8 @@ MModuleDEESMEX::MModuleDEESMEX() : MModule()
   
   // Default to adding noise to the sim data
   m_ApplyResolutionCalibration = true;
+  m_EnableShieldVeto = true;
+  m_EnableGuardRingVeto = true;
```
fhagemann, line 83: Just to confirm: we want `true` to be the default for `m_EnableShieldVeto` and `m_EnableGuardRingVeto`?

src/MModuleDEESMEX.cxx
```diff
@@ -194,7 +195,7 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
   // Step (10): Handles triggers and guard ring vetoes, pre-scalers, calculate dead-time, add nearest neighbor noise, calculate random coincidence time
   m_StripTrigger.Clear();
   m_StripTrigger.AnalyzeEvent(Event);
-  if (m_StripTrigger.HasGRVeto() == true) {
+  if (m_EnableGuardRingVeto == true && m_StripTrigger.HasGRVeto() == true) {
```
fhagemann: Maybe this just a personal preference, so feel free to ignore, but:
It would be nice if we were consistent in the naming `GuardRingVeto` and `GRVeto`.
In this line, we are mixing both conventions.

src/MSubModuleStripTrigger.cxx
```diff
@@ -535,6 +550,11 @@ bool MSubModuleStripTrigger::ParseDeadtimeFile()
   m_ASICDeadTimePerChannelFromFile = Parser.GetTokenizerAt(1)->GetTokenAtAsDouble(1);
   m_StripDelayAfter1FromFile = Parser.GetTokenizerAt(1)->GetTokenAtAsDouble(2);
   m_StripDelayAfter2FromFile = Parser.GetTokenizerAt(1)->GetTokenAtAsDouble(3);
+  if (Parser.GetTokenizerAt(1)->GetNTokens() > 4) {
+    m_FastClearDeadTimeFromFile = Parser.GetTokenizerAt(1)->GetTokenAtAsDouble(4);
+  } else {
+    m_FastClearDeadTimeFromFile = m_ASICDeadTimePerChannelFromFile;
```
ckierans: Why is this set to m_ASICDeadTimePerChannelFromFile if it's not provided? Do these just happen to be the same, or is there a physical motivation for why these are the same? I'm not sure how the FastClear parameter is defined/determined.

src/MSubModuleStripTrigger.cxx
```diff
@@ -129,6 +132,18 @@ void MSubModuleStripTrigger::Clear()
 ////////////////////////////////////////////////////////////////////////////////
 
 
+MTime MSubModuleStripTrigger::ApplyShieldVetoDeadtime(const MTime& ShieldVetoTime)
```
ckierans: I'm wondering if there's a better name for this function. It's not actually the shield deadtime that's being applied. It's raising the RTB enable before the shields even done reading out, right? Here you call that FastClear. Is that a hardware term folks are familiar with (it's new to me). Maybe "ApplyHardShieldVetoDeadtime" makes things a bit easier to understand? Or "ApplyHardShieldVetoFastClearDeadtime"... oof, that's getting long. Think about it.

include/MSubModuleStripTrigger.h
```diff
@@ -85,6 +85,9 @@ class MSubModuleStripTrigger : public MSubModule
   //! Return the time when the dead time ends - filled after AnalyzeEvent
   MTime GetDeadTimeEnd() const { return m_DeadTimeEnd; }
 
+  //! Apply strip deadtime caused by a shield veto
+  MTime ApplyShieldVetoDeadtime(const MTime& ShieldVetoTime);
```
ckierans: The description of this function here can be more clear. Maybe "Apply GeD deadtime cause by shield hard veto"

src/MModuleDEESMEX.cxx
```diff
@@ -160,11 +162,10 @@ bool MModuleDEESMEX::AnalyzeEvent(MReadOutAssembly* Event)
   // Step (6): the shield veto / trigger, handle pre-scalers, calculate dead-time, calculate random coincidence time
   m_ShieldTrigger.Clear();
   m_ShieldTrigger.AnalyzeEvent(Event);
-  if (m_ShieldTrigger.HasShieldVeto() == true) {
+  if (m_EnableShieldVeto == true && m_ShieldTrigger.HasShieldVeto() == true) {
     Event->SetShieldVeto(true);
-    if (m_ShieldTrigger.GetDeadTimeEnd() > m_DeadTimeEnd) {
-      m_DeadTimeEnd = m_ShieldTrigger.GetDeadTimeEnd();
-    }
+    MTime SelectedDeadTimeEnd = m_StripTrigger.ApplyShieldVetoDeadtime(m_ShieldTrigger.GetShieldVetoTime());
```
ckierans: This variable name isn't super informative. This is only the case when there's a hard veto and you're raising the enable again, right? Can you make that more clear vs "SelectedDeadTimeEnd"?

ckierans (commented):
Looks good, Parshad. Make sure you resolve the conflicts before its ready to merge.

While reviewing the code and thinking through the hard veto logic, I got a bit turned around with the m_DeadTimeEnd variable. There is the same variable name as a private member for both the shield triggers and the strip triggers, as well as the MModelDEESMEX class. I think this will become confusing/complicated once you start tracking the deadtime of the shields (for ACS event data) separately from the deadtime of the GeDs, since these are independant. m_DeadTimeEnd would then be different for the GeDs and the shields based on the number of triggered strips or shield channels, right? Perhaps just renaming the m_DeadTimeEnd variable as either GeD or Shield specific (i.e. m_GeDDeadTimeEnd) would clear this up for now.

src/MModuleDEESMEX.cxx
parshadkp: That's my understanding, but this is easy to change if needed.

src/MSubModuleStripTrigger.cxx
parshadkp: Thanks for catching this, it is a mistake. At first, I was setting "m_FastClearDeadTimeFromFile = m_ASICDeadTimePerChannelFromFile" since they are the same number but later I ended up introducing a new variable all together. Fixed in latest commit.

src/MModuleDEESMEX.cxx
parshadkp: The local variable isn't really needed because it is only used to update the m_DeadTimeEnd. Fixed in latest commit.

src/MSubModuleStripTrigger.cxx
parshadkp: How about ApplyFastClearDeadtime? I changed to this in the latest commit.

parshadkp:
Hi Carolyn, thanks for pointing out the m_DeadTimeEnd issue. Now DEESMEX module simply checks for the dead time directly from the respective trigger module. For strip trigger, this will return early if strip dead time is more than event time. For shield trigger, this will erase shield hits if the event is within shield dead time.

See example below:

if (HasStripHit == true && Event->GetTimeUTC() < m_StripTrigger.GetDeadTimeEnd()) {
    // Strip/GeD deadtime only gates the strip path. Shield processing above still runs.
    return true;
  }

src/MModuleDEESMEX.cxx
parshadkp: Fixed in latest commit.

resource/calibration/COSI_SingleDet/deadtimeParameters.txt
parshadkp: Fixed in latest commit.

resource/dee/deadtimeParameters.txt
```diff
@@ -0,0 +1,4 @@
+CoincidenceWindow (s); StripDeadtimePerStrip (s); delayAfter1 (s); delayAfter2 (s); FastClearDeadtime (s)
+1.4e-6		       1e-6	                  0.2e-6	   0.2e-6          1e-6
+ShieldDeadtimePerStrip (s); ShieldPulseDuration (s); ShieldDelayBefore (s); ShieldDelayAfter (s); ShieldVetoWindowDetla (s)
```
fhagemann:
```suggestion
ShieldDeadtimePerStrip (s); ShieldPulseDuration (s); ShieldDelayBefore (s); ShieldDelayAfter (s); ShieldVetoWindowDelta (s)
```

include/MModuleDEESMEX.h
```diff
@@ -165,6 +165,16 @@ class MModuleDEESMEX : public MModule
...
+  //! Enable or disable shield veto effects
+  bool GetEnableShieldVeto() const { return m_EnableShieldVeto; }
+  void SetEnableShieldVeto(bool EnableShieldVeto) {
+    m_EnableShieldVeto = EnableShieldVeto;
+  }
+  //! Enable or disable guard ring veto effects
+  bool GetEnableGuardRingVeto() const { return m_EnableGRVeto; }
+  void SetEnableGuardRingVeto(bool EnableGuardRingVeto) {
+    m_EnableGRVeto = EnableGuardRingVeto;
+  }
```
fhagemann: Now, you're still mixing `GR` and `GuardRing`.
As you're using `GuardRing` in most of the code, you might just want to swap all occurences of `GR` to `GuardRing` in this PR.

include/MSubModuleStripTrigger.h
```diff
@@ -83,7 +83,10 @@ class MSubModuleStripTrigger : public MSubModule
   bool HasGRVeto() const { return m_HasGRVeto; }
```
fhagemann: It looks like this is the only place where we use `GR` consistently, so maybe change it to `GuardRing` here (and everywhere where either `HasGRVeto` or `m_HasGRVeto` is used.

include/MModuleDEESMEX.h
parshadkp: All GR should now be GuardRing instead.

include/MSubModuleStripTrigger.h
parshadkp: All GR should be GuardRing instead.

src/MModuleDEESMEX.cxx
```diff
@@ -64,7 +64,7 @@ MModuleDEESMEX::MModuleDEESMEX() : MModule()
   
   // Allow the use of multiple threads and instances
   m_AllowMultiThreading = true;
-  m_AllowMultipleInstances = true;
+  m_AllowMultipleInstances = false;
```
fhagemann, line 67: Just to have this documented here: we set the multithreading back to false because it seems to slow down the DEE and it also results in the trigger statistics to be printed multiple times (once per instance)

fhagemann (approved):

src/MModuleDEESMEX.cxx
ckierans: Yes, this is standard operating mode.

ckierans (commented):
Looks good, Parshad!

PR #148, opened 2026-05-19
GuardRing Hits and XE in EVTA file

ckierans:
This is a draft PR that addresses issues with the position determination in the Depth Calibration. I presented at the CDEE meeting in April showing examples of these "bad" events in the evta file:

Screenshot 2026-05-18 at 11 35 31 PM

Andreas's comments for me at the meeting:
Screenshot 2026-05-18 at 11 37 48 PM
(Julian handed 5 already)

With this PR, I'm saving the GR hits in the evta file with the "GR" keyword and assigning them a random position in the guard ring volume. For events that fail the depth calibration, we're using the "XE" keyword and again just assign the hit a random position within the detector volume. These keywords are streamed to the evta file though the MHit::StreamEvta function, but I haven't yet done the same for the .dat files.

 Now, looking at the same events as above, we now get

> ID 999
> TI 1752086301.590697765
> HT 3;-1.45508;-0.403469;1.32867;365.288;0.0336036;0.0217617;0.0336036;0.961842
> CC NStripHits 3
> QA StripPairing (Best strip pairing leaves at least one grouping of strips unpaired (unpaired energy: 8.611698))
> PQ 1.17823 

Note this is a good event, and now we have the TI information from PR #124 , and better QA details from Strip pairing from PR #121 .

> SE
> ID 1000
> TI 1752086301.590716838
> XE 0.276994;0.3894;-2.94184;137.792
> XE 3.56288;0.753757;-0.862971;12.2433
> XE -2.24436;-0.563929;-1.24495;55.3574
> CC NStripHits 5
> BD DepthCalibrationError (Multiple hits on single strip)
> QA StripPairing (Event contains multiple hits on a single strip)
> PQ 0.386063

This event consisted of multiple hits on a single strip. It's only a QA flag in strip pairing since Julian is able to figure out the most likely ordering, but the depth calibration still fails to find the right position. Here these three hits are given the XE flag to track the extra energy deposit.

> SE
> ID 1092
> TI 1752086301.653759717
> HT 3;0.640234;-0.260901;-1.9307;240.615;0.0336036;0.0231942;0.0336036;0.93588
> XE 3.27438;-0.281045;1.91529;13.421
> GR -1.67307;0.345954;-3.75344;44.4647
> CC NStripHits 6
> BD DepthCalibrationError (Out of Range)
> BD GR Veto
> PQ 3.65235

[...]

PR #149, opened 2026-05-19
Adding the option to save/or not QA-flagged events

ckierans:
I cherry-picked this from PR #148 since it's much simpler to have this as it's own PR. 

I checked this with gse_20250709T113820.hdf for detector HP52358-3 and get 9.3 degrees when we allow QA-flagged events through, and 8.2 degrees when we don't. There's also a loss of ~50% of the events when remove QA-flagged hits. It seems like the majority of these hits that are filtered out with the QA flag are events when the best strip pairing leaves at least one grouping of strips unpaired.

@julianmgerber should compare with his ARM numbers to see if this selection recovers the better ARMs he was seeing before we let these lesser quality events through.

### DIFF
--- a/include/MGUIOptionsEventSaver.h
+++ b/include/MGUIOptionsEventSaver.h
@@ -81,6 +81,9 @@ class MGUIOptionsEventSaver : public MGUIOptions
   //! Checkbutton to save or reject bad events
   TGCheckButton* m_SaveBadEvents;
 
+  //! Checkbutton to save or reject quality flag events
+  TGCheckButton* m_SavePoorQualityEvents;
+
   //! Checkbutton to save veto events
   TGCheckButton* m_SaveVetoEvents;
 
--- a/include/MModuleEventSaver.h
+++ b/include/MModuleEventSaver.h
@@ -63,6 +63,11 @@ class MModuleEventSaver : public MModule
   //! Set whether the Bad events should be saved
   void SetSaveBadEvents(bool SaveBadEvents) { m_SaveBadEvents = SaveBadEvents; }
 
+  //! Return true if the Poor Quality events should be saved
+  bool GetSavePoorQualityEvents() const { return m_SavePoorQualityEvents; }
+  //! Set whether the Poor Quality events should be saved
+  void SetSavePoorQualityEvents(bool SavePoorQualityEvents) { m_SavePoorQualityEvents = SavePoorQualityEvents; }
+
   //! Return true if the Veto events should be saved
   bool GetSaveVetoEvents() const { return m_SaveVetoEvents; }
   //! Set whether the Veto events should be saved
@@ -192,6 +197,9 @@ class MModuleEventSaver : public MModule
   
   //! Save bad events
   bool m_SaveBadEvents;
+
+  //! Save poor quality events
+  bool m_SavePoorQualityEvents;
 
   //! Save Veto events
   bool m_SaveVetoEvents;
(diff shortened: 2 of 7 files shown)


julianmgerber:
I tried this out with data from 434-1 (gse_20260227T150331.hdf5). When I include QA-flagged events I get an ARM of 9.9 +/- 0.1 deg. When excluding QA-flagged events, I get 9.58 +/- 0.11 deg. However this doesn't improve the ARM all the way to the value I was getting during the original unit level calibration, 9.19 +/- 0.13 deg (which was done on March 6th). 

I also tried it on 434-2 data (gse_20260128T103501.hdf5). Again, the ARM I find when excluding QA-flagged events (9.77 +/- 0.16 deg) still doesn't get down to the older unit level calibration number (9.11 +/- 0.15 deg). I also tested a couple different slow thresholds to see if that could be playing a role. When I use a flat 20 keV cut, the ARM (excluding QA-flagged events) is 10.72 +/- 0.1 deg. When I use the actual per-strip calibration file (20260126_LBL_HP52434-2_slow_thresholds_avg.csv), I get a similar value of 10.32 +/- 0.13 deg. I'm not totally sure what to make of this since I would've expected the slow threshold to improve the ARM a bit.

ckierans:
Okay, thanks for checking. Let me see if I can track all of the QA and BD events that may have changed to understand why we're not recovering the same ARM with the QA filter.

ckierans:
Hmm, I'm seeing behavior that makes sense to me when looking at 358-3, including with the threshold file. @julianmgerber What Nuclearizer git commit were you using when you got the older unit level calibration numbers for the two detectors listed above? 

I checked out the develop/em branch from March '26 (commit 8114b7cd940b3e11defb78ea7c043617a3ee172e):
I get 7.7 +/- 0.3 degrees FWHM for the ARM fit of gse_20250709T113820.hdf when using the threshold file (vs the worse results listed above).

Then, checking out the this PR and keeping the QA events, I get 7.8 +/- 0.3 deg, aka consistent results with the analysis with the older commit. Excluding QA events get 7.1 +/- 0.3 degrees. Again, these both are using the threshold file, I'm ignoring NN strip hits, and I'm using your strip pairing algorithm. 

For me, using a flat 20 keV threshold gives worse results than the file threshold (7.7 deg vs 7.1 deg). I think this is because cutting with 20 keV is quite a bit larger than the thresholds in the file, so we're missing information and therefore getting worse reconstruction for some events (the spectrum shows a large low-energy tail with the 20 keV threshold). When you say you're expecting the slow thresholds to improve the ARM, do you mean you expected a better improvement with the file, or you expected a better improvement with the 20 keV cut?

julianmgerber:
@ckierans I'll try running it for 358-3 as well and see if I get the same numbers just to make sure I'm not doing anything wrong. For the original 434-1 unit level calibration, I believe I was on commit 3dd7d93. For 434-2, I originally ran it on a branch that had the updated strip pairing but was otherwise a bit behind develop/em. Essentially, I believe it was commit 2cd0898 with 8eb29cd cherry-picked. 

> For me, using a flat 20 keV threshold gives worse results than the file threshold (7.7 deg vs 7.1 deg). I think this is because cutting with 20 keV is quite a bit larger than the thresholds in the file, so we're missing information and therefore getting worse reconstruction for some events (the spectrum shows a large low-energy tail with the 20 keV threshold). When you say you're expecting the slow thresholds to improve the ARM, do you mean you expected a better improvement with the file, or you expected a better improvement with the 20 keV cut?

I think I expected the ARM to improve with the threshold cut file compared to no threshold cut. But the ARM I'm getting with the threshold cut file (10.32 deg) is worse than with no threshold cut at all (9.77 deg)

PR #152, opened 2026-05-21
Set default value of `m_Coeffs_Energy` in depth calibration to 59.5keV

fhagemann:
In `MModuleDepthCalibration`, the reference energy at which the depth calibration was performed is stored as `m_Coeffs_Energy` and expected to be read from the header of the depth coefficients file.

https://github.com/cositools/nuclearizer/blob/d153b0fea708f993e514e1b8a204128b992b16ec/src/MModuleDepthCalibration.cxx#L485-L486
https://github.com/cositools/nuclearizer/blob/d153b0fea708f993e514e1b8a204128b992b16ec/src/MModuleDepthCalibration.cxx#L497-L501

However, most (all?) of the depth calibration coefficient files do not have that header, resulting in `m_Coeffs_Energy` to stay zero and resulting a scaled `noiseFWHM` of 0:
https://github.com/cositools/nuclearizer/blob/d153b0fea708f993e514e1b8a204128b992b16ec/src/MModuleDepthCalibration.cxx#L462-L477

While the zero `noiseFWHM` is handled in the forward pipeline by just setting a "standard" FWHM of 3keV, this will result in wrong results in the DEE inverse depth cal in PR #134. And I mean: we have the information on the FWHM in the files.

As we're currently ALWAYS generating the depth calibration coefficients at 59.5keV, I propose to update that default values, such that people can still use the old files that might not have the headers in them.

### DIFF
--- a/src/MModuleDepthCalibration.cxx
+++ b/src/MModuleDepthCalibration.cxx
@@ -84,7 +84,7 @@ MModuleDepthCalibration::MModuleDepthCalibration() : MModule()
   m_AllowMultiThreading = true;
   m_AllowMultipleInstances = false;
 
-  m_Coeffs_Energy = 0;
+  m_Coeffs_Energy = 59.5;
 
   m_NoError = 0;
   m_Error1 = 0;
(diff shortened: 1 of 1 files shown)


ckierans:
Thanks, Felix!

PR #153, opened 2026-05-23
[DEE] Allow multithreading and multiple instances in the DEE

fhagemann:
So far, setting `m_AllowMultipleInstances` to `true` in `MModuleDEESMEX`,  would cause a segmentation fault.
This was caused by `m_Geometry` not being set and containing uninitialized garbage.

The root cause for this is how megalib handles creating multiple instances (find the code [here](https://github.com/zoglauer/megalib/blob/31298310d572e23a348c731eda0b8637abebde7a/src/fretalon/framework/src/MSupervisor.cxx#L982-L999)):

- It creates a new instance by cloning the module via `Clone()`
- It then checks if the module can be initialized using `Initialize()`

In the case of `MModuleDEESMEX`, when cloning the module, we don't pass `m_Geometry` down to the clone, and then get weird behavior when calling on uninitialized garbage. 

To avoid this, in this PR:
- we set `m_Geometry` in the constructor of `MModuleDEESMEX` to `nullptr`
- check in `Initialize` if it is `nullptr`, or whether it was actually set
- pass the geometry to clones of `MModuleDEESMEX` in `Clone()`
- enable running the DEE with multiple instances

It seems to work (and be significantly faster than before) on my laptop, but feedback on this is very much appreciated!

### DIFF
--- a/include/MModuleDEESMEX.h
+++ b/include/MModuleDEESMEX.h
@@ -55,7 +55,9 @@ class MModuleDEESMEX : public MModule
   //! Create a new object of this class
   virtual MModuleDEESMEX* Clone()
   {
-    return new MModuleDEESMEX();
+    MModuleDEESMEX* M = new MModuleDEESMEX();
+    M->SetGeometry(m_Geometry);
+    return M;
   }
 
   //! Set the geometry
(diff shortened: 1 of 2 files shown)


fhagemann:
Side note: I think this might also be relevant for other modules relying on `m_Geometry`, e.g. `MModuleDepthCalibration`.

zoglauer:
Looks reasonable. Approved to merge.

PR #156, opened 2026-05-27
Stream crystal hits to "UC" instead of "UH" in ROA output

fhagemann:
Currently, when running simulations with the EM massmodels, the current implementation of the DEE results in `MStripHits` and `MCrystalHits` both being saved to `UH` to the ROA file. This messes up parsing these ROA files, because the rows for crystal hits have fewer entries:
Screenshot from 2026-05-27 09-15-01

I was wondering whether `UH` here is actually correct, or whether it should be exclusively reserved for `MStripHits`.
It looks like `MShieldCrystalHits` for example are saved to `UC` instead of `UH`:
https://github.com/cositools/nuclearizer/blob/79ac09121379fdbfab2389c1effd2ebc4232ea1a/src/MShieldCrystalHit.cxx#L159-L163

whereas `MCrystalHits` are saved to `UH`:
https://github.com/cositools/nuclearizer/blob/79ac09121379fdbfab2389c1effd2ebc4232ea1a/src/MCrystalHit.cxx#L159-L163

Writing `MCrystalHits` also to `UC` instead of `UH` would prevent my current scripts from failing in case ACS hits are present in the simulations :innocent:

### DIFF
--- a/src/MCrystalHit.cxx
+++ b/src/MCrystalHit.cxx
@@ -160,7 +160,7 @@ void MCrystalHit::StreamRoa(ostream& S, bool WithADC, bool WithEnergy, bool With
 {
   //! Stream the content in MEGAlib's evta format 
 
-  S<<"UH "
+  S<<"UC "
    <<m_ReadOutElement->GetDetectorID()<<" ";
   if (WithADC == true) {
     S<<m_ADCUnits<<" ";


zoglauer:
Let me think about this - the roa / UH format is the universal format to get stuff into melinator.

zoglauer:
The dat format was originally intended to just store all the information which was in the class.

zoglauer:
we just stopped updating it at some point in time.

fhagemann:
Also, `MShieldCrystalHit` also looks like an unused left-over..

ckierans:
@zoglauer Any more thoughts on using "UC" instead of "UH" in the roa file for shield hits? This has been an open PR for awhile now.

fhagemann:
@zoglauer I guess this now needs to be closed / revisited with the newest changes in megalib?

zoglauer:
Yes, I already have a working version with all the changes on my Mac - we will have a new pull request soon.

fhagemann:
I'm excited :)

PR #158, opened 2026-05-29
[DEE] Update inverse depth calibration from pixel-based to strip-based

fhagemann:
Currently, we apply stretch and offset to the CTD/TAC values on a pixel by pixel basis.
Field looked into composing these stretches and offsets on a strip basis, and this PR provides the nuclearizer code to read it.

The file format currently looks like this (very preliminary, this might still change depending if we need more information):
```
#detector_info,detector_id,detector_name,depth_spline_filename,mean_stretch,mean_offset
#detector_info,0,lbl_HP52406-3,HV32_LV32_59.5keV_lbl_HP52406-3_15.93mm_1050V_4.53e9_83K_ctd_depth_mcSPICE,1.124962607911977,21.15978945295241
#
#STRIP_ID,detector,side,strip,stretch_correction,offset_correction,timing_resolution_1σ_60_keV
96,0,0,0,1.0460236944724564,-9.128100617278196,9.892898259233435
97,0,0,1,1.0290178925959224,-8.802895822507075,8.486136187899895
98,0,0,2,1.021702711739791,-10.675809115071125,7.525486207481912
99,0,0,3,1.0120196046486603,-9.652688271376434,8.092190648673894
100,0,0,4,1.005529715566832,-6.459132565872501,8.476423433322116
101,0,0,5,1.0012518297022444,-8.861778237543035,8.393700918283095
102,0,0,6,1.0057693308480216,-9.151225747582288,8.141965836063068
...
```

I made some comparison to show that the pixel-based and strip-based files yield similar results (406-3 in STTC, illuminating with 241Am from the HV side):

image

### DIFF
--- a/src/MModuleDepthCalibration.cxx
+++ b/src/MModuleDepthCalibration.cxx
@@ -538,6 +538,67 @@ bool MModuleDepthCalibration::LoadCoeffsFile(MString FileName)
 /////////////////////////////////////////////////////////////////////////////////
 
 
+bool MModuleDepthCalibration::LoadStripCoeffsFile(MString FileName)
+{
+
+  MFile CoeffsFile;
+  if (CoeffsFile.Open(FileName) == false) {
+    cout << "ERROR in MModuleDepthCalibration::LoadStripCoeffsFile: failed to open strip coefficients file." << endl;
+    return false;
+  }
+
+  MString Line;
+  while (CoeffsFile.ReadLine(Line) == true) {
+    if (Line.BeginsWith('#') == true) {
+      std::vector<MString> Tokens = Line.Tokenize(",");
+      if (Tokens.size() < 5 || Tokens[0] != "#detector_info" || Tokens[1] == "detector_id") continue;
+
+      int DetID = Tokens[1].ToInt();
+      // MString DetectorName = Tokens[2];
+      // MString DepthSplineFileName = Tokens[3];
+      m_MeanStretch[DetID] = Tokens[4].ToDouble();
+      m_MeanOffset[DetID] = Tokens[5].ToDouble();
+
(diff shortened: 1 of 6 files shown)


ckierans:
Do we want to provide the option to provide a strip-based calibration or a pixel-based calibration? Or are we moving entirely to a strip-based calibration?

fhagemann:
I mean, the code exists for both, so we could also allow for either.
I would favor strip-based calibration though, because with pixel-based calibrations we might get very weird artifacts of TAC values for hits on the same strip being inconsistent (e.g. if neighboring pixels on the same strip have different stretch/offsets).

ckierans:
So we'll need to make the same changes to the forward pipeline Depth Calibration module.

fhagemann:
My understanding was to keep the pixel-based calibration for the forward pipeline, but have the strip-based calibration for the DEE. Let's discuss this together with Field in tomorrow's meeting.

fhagemann:
I would postpone this PR to after having merged #165, as the latter might change where things are calculated/stored.

src/MModuleDepthCalibration.cxx
```diff
@@ -538,6 +538,67 @@ bool MModuleDepthCalibration::LoadCoeffsFile(MString FileName)
 /////////////////////////////////////////////////////////////////////////////////
 
 
+bool MModuleDepthCalibration::LoadStripCoeffsFile(MString FileName)
```
fhagemann, line 541: Up for debate:
I added the `LoadStripCoeffsFile` to read the strip-based depth coefficients file to `MModuleDepthCalibration`, even though it is never used in the forward pipeline.
The thought behind this was that both `MSubModuleChargeTransport` and `MSubModuleDepthReadout` require reading this file for z->drift time and timing smearing, respectively.

PR #160, opened 2026-06-03
Update reference TRA files

fhagemann:
When pulling the latest changes from `develop/em`, it looks like the end-to-end unit tests fail.
I tracked this down to #152, which now correctly reads the timing noise from the depth coefficients file to estimate the uncertainty. This influences:
- the uncertainty propagated to `PP` in photo-absorption events
  image
- the uncertainty propagated to `CD` (Compton interactions) and the resulting lever arm `LA`
  image
- some low-energy events close to the detector faces actually passing the depth calibration now, because the bigger FWHM tolerance allows to keep them 
  image

This PR updates the reference TRA files accordingly, so that the unit tests pass again


fhagemann:
Actually, it seems like it's not changing the uncertainty (I mixed that up with the `PP` column of a `MPairEvent`, where the second number is the energy uncertainty), but the `Y`-Position, so we're actually getting different results from the depth calibration now..

fhagemann:
Ok, the depth is actually calculated using probability weighting, so I guess these changes are expected with different timing noise now.

https://github.com/cositools/nuclearizer/blob/ab932fdee9c1578fcd205f9086ead87ddb53ee21/src/MModuleDepthCalibration.cxx#L315-L357

zoglauer:
From the unit test perspective it looks OK. You can merge it.

zoglauer (approved):

PR #161, opened 2026-06-05
CHG: Unit test changes for MTime assuming all times are UTC

zoglauer:
Attached are the changes required for the updated MTime class assuming all times are UTC.


fhagemann:
How is this different from my second commit in #132 ?

fhagemann (approved):
Looks good to me. I don't see the respective change in `megalib:develop-cosi` though (only on `megalib:main` so far)

zoglauer:
I forgot to push the changes to megalib:develop-cosi.

fhagemann:
I actually had to make a tiny correction to one of the `CT` lines in the reference TRA file in #132 to make this work again (difference is in the last digit)..

PR #164, opened 2026-06-09
CHG: Handle numerical accuracy issues

zoglauer:
MEGAlib's unit test class now contains a new function, which compares two files for numerical identity, i.e., allows for the last digit in numbers to be off.

This PR applies this to the unit tests.

Please update MEGAlib first.


fhagemann:
I can confirm that the tests that failed BEFORE this PR now pass AFTER this PR. Thanks for the quick fix!

Issue #111, opened 2026-03-23
Combining strip hits in the DEE


fhagemann:
Right now, strip hits are combined in the charge-transport DEE module, causing us to lose information from cosima, e.g. simulated position, which might be needed later in the DEE depth simulation module.

We might need to move the merging of strip hits to later in the pipeline to avoid losing the information where it is needed.

fhagemann:
This issue was resolved with #165 by moving the determination of drift times from the interaction depth already to `MSubModuleChargeTransport`.

PR #165, opened 2026-06-15
[DEE] Add drift times from depth calibration already in `MSubModuleChargeTransport`

fhagemann:
This PR addresses issue #111, and is a first step to add realistic nearest-neighbor timing to the DEE.

TLDR: I'm moving applying the depth splines + stretch + offset from `MSubModuleDepthReadout` to `MSubModuleChargeTransport`, while keeping the smearing of timing values in `MSubModuleDepthReadout`,
while adding an extra field to `MDEEStripHit` to keep track of the expected time at which the fast-shaper signal peaks.

Some background information:
- Scenario 1: an event where all charge is collected on one strip, the **main strip**.
  - The fast-shaper signal of the main strip is unipolar, with a very pronounced peak. We use that for timing and depth calibration.
  - The nearest-neighbor (spectator) strips typically feature a bipolar fast-shaper. The fast-shaper signal for these strips features a peak-dip structure that peaks ~50ns earlier than the main strip (for most of the detector depth, ignoring events VERY close to the HV and LV sides)
  Screenshot from 2026-06-15 13-50-19

- Scenario 2: an event where the charge is shared between two neighboring strips.
  - Here, the fast-shaper of both collecting strips is a superposition of the above-mentioned unipolar (direct) signal, and the bipolar (transient) signal, resulting both of them to peak earlier compared the direct signal in scenario 1.
  Screenshot from 2026-06-15 13-48-11

To first order, the relation between when the fast-shaper signal peaks and how much energy it records is pretty linear, i.e. `t_drift = t_unipolar - 50ns * (1 - E_recorded / E_event)`

The drift times for the **main** strips can be determined from the simulated electron and hole drift times, by applying stretch and offset from the depth calibration. For main strips with no charge sharing, the "time until fast shaper peaks" should be similar to those drift times.

In order to include simulated drift times in the DEE, we need to know which strips are main strips / NN strips and if there was charge sharing present or not, because for these the "time until fast shaper peaks" is not identical to the charge drift times, and we need additional information (basically the depth, `Z` and if/how much energy was charge-shared), all of which is only available in `MSubModuleChargeTransport`. Therefore, the determination of charge drift times + stretch + offset is moved to `MSubModuleChargeTransport`, and for now we apply nearest-neighbor timing by using the simplistic linear relationship `t_drift = t_unipolar - 50ns * (1 - E_recorded / E_event)`.

### DIFF
--- a/include/MDEEStripHit.h
+++ b/include/MDEEStripHit.h
@@ -88,6 +88,8 @@ struct MDEEStripHit
   unsigned int m_TAC;
   //! The calibrated timing in ns
   double m_Timing;
+  //! The drift time in ns since creation of event
+  double m_DriftTime;
   //! The measured temperature value
   double m_Temperature;
 
--- a/include/MDEEStripHit.h
+++ b/include/MDEEStripHit.h
@@ -84,6 +84,8 @@ struct MDEEStripHit
   unsigned int m_ADC;
   //! If the strip exceeds thresholds
   bool m_HasTriggered;
+  //! If the strip has fast timing
+  bool m_HasFastTiming;
   //! The measured TAC value in ADC units
   unsigned int m_TAC;
   //! The calibrated timing in ns
--- a/include/MDEEStripHit.h
+++ b/include/MDEEStripHit.h
@@ -88,8 +88,8 @@ struct MDEEStripHit
   unsigned int m_TAC;
   //! The calibrated timing in ns
   double m_Timing;
-  //! The drift time in ns since creation of event
-  double m_DriftTime;
+  //! The fast shaper peaking time in ns since creation of event
+  double m_FastPeakTime;
   //! The measured temperature value
   double m_Temperature;
 
--- a/src/MSubModuleChargeTransport.cxx
+++ b/src/MSubModuleChargeTransport.cxx
@@ -250,7 +250,7 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     while (IterLV2 != LVHits.end()) {
       if (IterLV1->m_ROE == IterLV2->m_ROE) {
         IterLV1->m_Energy += IterLV2->m_Energy;
-        IterLV1->m_DriftTime = IterLV1->m_Energy > IterLV2->m_Energy ? IterLV1->m_DriftTime : IterLV2->m_DriftTime;
+        IterLV1->m_FastPeakTime = IterLV1->m_Energy > IterLV2->m_Energy ? IterLV1->m_FastPeakTime : IterLV2->m_FastPeakTime;
         IterLV2 = LVHits.erase(IterLV2);
       } else {
         ++IterLV2;
@@ -262,7 +262,7 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     while (IterHV2 != HVHits.end()) {
       if (IterHV1->m_ROE == IterHV2->m_ROE) {
         IterHV1->m_Energy += IterHV2->m_Energy;
-        IterHV1->m_DriftTime = IterHV1->m_Energy > IterHV2->m_Energy ? IterHV1->m_DriftTime : IterHV2->m_DriftTime;
+        IterHV1->m_FastPeakTime = IterHV1->m_Energy > IterHV2->m_Energy ? IterHV1->m_FastPeakTime : IterHV2->m_FastPeakTime;
         IterHV2 = HVHits.erase(IterHV2);
       } else {
         ++IterHV2;
@@ -339,8 +339,8 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
   if (ID >= 0 && ID < NStrips && std::abs(Q) <= QWidth/2.0 && std::hypot(P, Q) <= Radius) {
 
     // Determine the charge drift times in nanoseconds from simulations + stretch/offset from the depth calibration
-    // Set the default to 0ns in case no depth calibration coefficients exist
-    double DriftTime = 1e10;
+    // Set the default to a large number (here: 1e10 ns) in case no depth calibration coefficients exist
+    double FastPeakTime = 1e10;
 
     TSpline3* DriftTimeSpline = isLV ? m_HoleDriftSplines[DetID] : m_ElectronDriftSplines[DetID];
     int PixelCode = 10000*DetID + 100*(isLV ? ID : OppositeStripID) + (isLV ? OppositeStripID : ID);
@@ -352,12 +352,11 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
       const vector<double>& Coeffs = it->second;
       double Stretch = Coeffs[0];
       double Offset = isLV ? Coeffs[1] : 0.0;
-      DriftTime = (DriftTimeSpline->Eval(Z) + Offset) * Stretch;
+      FastPeakTime = (DriftTimeSpline->Eval(Z) + Offset) * Stretch;
     } else {
       if (g_Verbosity >= c_Warning) {
         cout << "No depth calibration coefficients for pixel in DetID " << DetID << " HV " << (isLV ? OppositeStripID : ID) << " LV " << (isLV ? ID : OppositeStripID) << endl;
       }
-      DriftTime = 1e10;
     }
 
     // Apply charge sharing based on relative coordinate to the gap of that strip (0 <= X < XPitch)
@@ -387,7 +386,7 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
     MainSH.m_ROE.SetStripID(ID);
     MainSH.m_OppositeStripID = OppositeStripID;
     MainSH.m_Energy = MainStripEnergy;
-    MainSH.m_DriftTime = DriftTime - 50 * (1 - MainStripEnergy / SH.m_SimulatedEnergy);
+    MainSH.m_FastPeakTime = FastPeakTime - 50 * (1 - MainStripEnergy / SH.m_SimulatedEnergy);
     MainSH.m_IsGuardRing = false;
     m_ChargeTransportHits.push_back(MainSH);
 
@@ -395,7 +394,7 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
     // if (NNLeftStripEnergy > IonizationEnergy) {
       MDEEStripHit NNLeftSH = SH;
       NNLeftSH.m_Energy = std::max(NNLeftStripEnergy, 0.0);
-      NNLeftSH.m_DriftTime = DriftTime - 50 * (1 - NNLeftStripEnergy / SH.m_SimulatedEnergy);
+      NNLeftSH.m_FastPeakTime = FastPeakTime - 50 * (1 - NNLeftStripEnergy / SH.m_SimulatedEnergy);
       NNLeftSH.m_OppositeStripID = OppositeStripID;
       if (ID > 0) {
         NNLeftSH.m_ROE.SetStripID(ID - 1);
@@ -411,7 +410,7 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
     // if (NNRightStripEnergy > IonizationEnergy) {
       MDEEStripHit NNRightSH = SH;
       NNRightSH.m_Energy = std::max(NNRightStripEnergy, 0.0);
-      NNRightSH.m_DriftTime = DriftTime - 50 * (1 - NNRightStripEnergy / SH.m_SimulatedEnergy);
+      NNRightSH.m_FastPeakTime = FastPeakTime - 50 * (1 - NNRightStripEnergy / SH.m_SimulatedEnergy);
       NNRightSH.m_OppositeStripID = OppositeStripID;
       if (ID < NStrips - 1) {
         NNRightSH.m_ROE.SetStripID(ID + 1);
--- a/src/MSubModuleChargeTransport.cxx
+++ b/src/MSubModuleChargeTransport.cxx
@@ -394,36 +394,32 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
     m_ChargeTransportHits.push_back(MainSH);
 
     // create MDEEStripHit for the left NN
-    // if (NNLeftStripEnergy > IonizationEnergy) {
-      MDEEStripHit NNLeftSH = SH;
-      NNLeftSH.m_Energy = std::max(NNLeftStripEnergy, 0.0);
-      NNLeftSH.m_FastPeakTime = FastPeakTime - 50 * (1 - NNLeftStripEnergy / SH.m_SimulatedEnergy);
-      NNLeftSH.m_OppositeStripID = OppositeStripID;
-      if (ID > 0) {
-        NNLeftSH.m_ROE.SetStripID(ID - 1);
-        NNLeftSH.m_IsGuardRing = false;
-      } else {
-        NNLeftSH.m_ROE.SetStripID(NStrips);
-        NNLeftSH.m_IsGuardRing = true;
-      }
-      m_ChargeTransportHits.push_back(NNLeftSH);
-    // }
+    MDEEStripHit NNLeftSH = SH;
+    NNLeftSH.m_Energy = std::max(NNLeftStripEnergy, 0.0);
+    NNLeftSH.m_FastPeakTime = FastPeakTime - 50 * (1 - NNLeftStripEnergy / SH.m_SimulatedEnergy);
+    NNLeftSH.m_OppositeStripID = OppositeStripID;
+    if (ID > 0) {
+      NNLeftSH.m_ROE.SetStripID(ID - 1);
+      NNLeftSH.m_IsGuardRing = false;
+    } else {
+      NNLeftSH.m_ROE.SetStripID(NStrips);
+      NNLeftSH.m_IsGuardRing = true;
+    }
+    m_ChargeTransportHits.push_back(NNLeftSH);
     
     // create MDEEStripHit for the right NN
-    // if (NNRightStripEnergy > IonizationEnergy) {
-      MDEEStripHit NNRightSH = SH;
-      NNRightSH.m_Energy = std::max(NNRightStripEnergy, 0.0);
-      NNRightSH.m_FastPeakTime = FastPeakTime - 50 * (1 - NNRightStripEnergy / SH.m_SimulatedEnergy);
-      NNRightSH.m_OppositeStripID = OppositeStripID;
-      if (ID < NStrips - 1) {
-        NNRightSH.m_ROE.SetStripID(ID + 1);
-        NNRightSH.m_IsGuardRing = false;
-      } else {
-        NNRightSH.m_ROE.SetStripID(NStrips);
-        NNRightSH.m_IsGuardRing = true;
-      }
-      m_ChargeTransportHits.push_back(NNRightSH);
-    // }
+    MDEEStripHit NNRightSH = SH;
+    NNRightSH.m_Energy = std::max(NNRightStripEnergy, 0.0);
+    NNRightSH.m_FastPeakTime = FastPeakTime - 50 * (1 - NNRightStripEnergy / SH.m_SimulatedEnergy);
+    NNRightSH.m_OppositeStripID = OppositeStripID;
+    if (ID < NStrips - 1) {
+      NNRightSH.m_ROE.SetStripID(ID + 1);
+      NNRightSH.m_IsGuardRing = false;
+    } else {
+      NNRightSH.m_ROE.SetStripID(NStrips);
+      NNRightSH.m_IsGuardRing = true;
+    }
+    m_ChargeTransportHits.push_back(NNRightSH);
 
   } else {
     // TODO: implement charge sharing also for GR events
--- a/src/MSubModuleChargeTransport.cxx
+++ b/src/MSubModuleChargeTransport.cxx
@@ -34,9 +34,11 @@
 
 // MEGAlib libs:
 #include "MSubModule.h"
-#include "MModuleDepthCalibration.h"
 #include "MDShapeIntersection.h"
 #include "MDShapeTUBS.h"
+
+// Nuclearizer libs:
+#include "MModuleDepthCalibration.h"
 
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -390,9 +392,9 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool 
     m_ChargeTransportHits.push_back(MainSH);
 
     // create MDEEStripHit for the left NN
-    if (NNLeftStripEnergy > IonizationEnergy) {
+    // if (NNLeftStripEnergy > IonizationEnergy) {
       MDEEStripHit NNLeftSH = SH;
-      NNLeftSH.m_Energy = NNLeftStripEnergy;
+      NNLeftSH.m_Energy = std::max(NNLeftStripEnergy, 0.0);
       NNLeftSH.m_DriftTime = DriftTime - 50 * (1 - NNLeftStripEnergy / SH.m_SimulatedEnergy);
       NNLeftSH.m_OppositeStripID = OppositeStripID;
       if (ID > 0) {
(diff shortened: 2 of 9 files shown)


include/MDEEStripHit.h
```diff
@@ -88,6 +88,8 @@ struct MDEEStripHit
   unsigned int m_TAC;
   //! The calibrated timing in ns
   double m_Timing;
+  //! The drift time in ns since creation of event
+  double m_DriftTime;
```
fhagemann: Maybe the name `m_DriftTime` might not be correct.
It is essentially the time between the creation of the electron-hole pairs until the fast shaper on a non-charge-sharing main strip would peak, minus some offset accounting for the delay in peaking... 
For non-charge-sharing main strips, this value should correspond to the drift time, but it's not the case for charge-sharing strips or nearest neighbors. Suggestions welcome ^^

src/MSubModuleChargeTransport.cxx
```diff
@@ -195,6 +250,7 @@ bool MSubModuleChargeTransport::AnalyzeEvent(MReadOutAssembly* Event)
     while (IterLV2 != LVHits.end()) {
       if (IterLV1->m_ROE == IterLV2->m_ROE) {
         IterLV1->m_Energy += IterLV2->m_Energy;
+        IterLV1->m_DriftTime = IterLV1->m_Energy > IterLV2->m_Energy ? IterLV1->m_DriftTime : IterLV2->m_DriftTime;
```
fhagemann: The way we combine timing values when merging strip hits might not be too accurate. This works well for hits that were initially close together, but might be wrong for events that are spatially more separated.
If the fast-shaper direct signals would feature two separated peaks (usually more than 50ns apart), then it would depend on if the first peak would be above threshold etc. ... 
Maybe this merging can only happen during `MSubModuleDepthReadout`, when we know what the fast threshold is?

src/MSubModuleChargeTransport.cxx
```diff
@@ -280,13 +338,35 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
...
+
     // Apply charge sharing based on relative coordinate to the gap of that strip (0 <= X < XPitch)
     double FromGap = std::fmod(P + PWidth/2.0, PPitch);
 
     // Charge transport based on Eq. (7) in https://doi.org/10.1016/j.nima.2023.168310
     // calculate σ and η, assuming t = z / v = z / (µ * E)
-    double Sigma = std::sqrt(2.0 * kB * Temperature * ΔZ / (ElementaryCharge * MeanElectricField)); // in cm
-    double Eta   = std::cbrt(std::pow(InitialChargeCloudSize, 3) + 3.0 * N * ElementaryCharge * ΔZ / (4.0 * TMath::Pi() * Epsilon0 * EpsilonR * MeanElectricField)); // in cm
+    double Sigma = std::sqrt(2.0 * kB * Temperature * DeltaZ / (ElementaryCharge * MeanElectricField)); // in cm
+    double Eta   = std::cbrt(std::pow(InitialChargeCloudSize, 3) + 3.0 * N * ElementaryCharge * DeltaZ / (4.0 * TMath::Pi() * Epsilon0 * EpsilonR * MeanElectricField)); // in cm
```
fhagemann, line 370: Here, we could also think about updating `t = z / v = z / (µ * E)` to using the simulated drift times. Two caveats though: the files have those simulated drift times already convolved with electronics, and this might be quite sensitive to how you apply the depth-calibration offset (e.g. adding it to the hole drift times or subtracting it from the electron hole drift times)

src/MSubModuleChargeTransport.cxx
```diff
@@ -307,40 +387,41 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
     MainSH.m_ROE.SetStripID(ID);
     MainSH.m_OppositeStripID = OppositeStripID;
     MainSH.m_Energy = MainStripEnergy;
+    MainSH.m_DriftTime = DriftTime - 50 * (1 - MainStripEnergy / SH.m_SimulatedEnergy);
     MainSH.m_IsGuardRing = false;
     m_ChargeTransportHits.push_back(MainSH);
 
     // create MDEEStripHit for the left NN
-    if (NNLeftStripEnergy > IonizationEnergy) {
+    // if (NNLeftStripEnergy > IonizationEnergy) {
```
fhagemann: Here, I'm keeping all events, no matter if their energy is 0, just to keep the nearest-neighbor timing information.

src/MSubModuleChargeTransport.cxx
```diff
@@ -307,40 +387,41 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
     MainSH.m_ROE.SetStripID(ID);
     MainSH.m_OppositeStripID = OppositeStripID;
     MainSH.m_Energy = MainStripEnergy;
+    MainSH.m_DriftTime = DriftTime - 50 * (1 - MainStripEnergy / SH.m_SimulatedEnergy);
```
fhagemann: This is how to account for the difference in timing with respect to having a non-charge-sharing main strip. I know that this implementation is very simplistic, but captures what we see well for most events that are not close to the HV/LV sides.

We have one of our undergrads (Isidro) looking into this is more detail.
I would push going for higher-order implementations in a future PR.

src/MSubModuleDepthReadout.cxx
```diff
@@ -164,211 +132,105 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
+    if (SH.m_IsGuardRing == false) {
+      unsigned int StripID = SH.m_ROE.GetStripID();
+      if (SH.m_DriftTime > -200.0 ) {
+        SH.m_Timing = 4200.0 - SH.m_DriftTime;
+        if (m_ApplyTimingResolutionCalibration == true){
+          int PixelCode = 10000*DetID + 100*StripID + SH.m_OppositeStripID;
+          if (m_Coeffs.count(PixelCode) == 1){
+            vector<double> Coeffs = m_Coeffs[PixelCode];
+            double CTD_FWHM = Coeffs[2] * m_Coeffs_Energy / SH.m_Energy;
+            double CTD_Sigma = CTD_FWHM / 2.355;
```
fhagemann, line 153: `MSubModuleDepthReadout` now only smears the final timing values and applies the inverse TAC calibration. Everything else is done in `MSubModuleChargeTransport` now.

src/MSubModuleChargeTransport.cxx
```diff
@@ -280,13 +338,35 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
...
+    if (it != m_Coeffs.end()) {
+      const vector<double>& Coeffs = it->second;
+      double Stretch = Coeffs[0];
+      double Offset = isLV ? Coeffs[1] : 0.0;
+      DriftTime = (DriftTimeSpline->Eval(Z) + Offset) * Stretch;
+    } else {
+      if (g_Verbosity >= c_Warning) {
+        cout << "No depth calibration coefficients for pixel in DetID " << DetID << " HV " << (isLV ? OppositeStripID : ID) << " LV " << (isLV ? ID : OppositeStripID) << endl;
+      }
+      DriftTime = 1e10;
```
fhagemann: If we don't have stretch and offset available, I'm setting the drift times unreasonably high. Using a pixel-based inverse depth calibration, this can result in:
- if there is a HV strip without fast timing, all events collected on that strip would also give unreasonable drift times on the LV strips.
- If we were to use these charge drift times for diffusion/self-repulsion, this would screw the calculation up and result in gigantic amounts of charge sharing.

This would be avoided by transitioning from pixel-based to strip-based inverse depth calibration (see #158).

I'm also open to suggestions how to deal with events with no depth calibration.

src/MSubModuleDepthReadout.cxx
```diff
@@ -164,211 +132,105 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
-          // Convert drift time to timing by subtracting 4200ns (for now)
-          // TODO: Improve determining the timing from drift times
-          SH.m_Timing = 4200.0 - ElectronDriftTime;
-
-          // Smear the timing value based on the given CTD resolution
-          // --> divide by √2 to obtain TAC resolution from CTD resolution
-          if (m_ApplyTimingResolutionCalibration == true) {
+    if (SH.m_IsGuardRing == false) {
+      unsigned int StripID = SH.m_ROE.GetStripID();
+      if (SH.m_DriftTime > -200.0){
```
fhagemann: This is an arbitrary number right now.
This number CAN be negative (if the charge drift time is shorter than 50ns, and the neighbor did not charge share, we might get e.g. 40ns for the main strip and -10ns for its nearest neighbor).

src/MSubModuleChargeTransport.cxx
```diff
@@ -307,40 +387,41 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
...
+    MainSH.m_DriftTime = DriftTime - 50 * (1 - MainStripEnergy / SH.m_SimulatedEnergy);
     MainSH.m_IsGuardRing = false;
     m_ChargeTransportHits.push_back(MainSH);
 
     // create MDEEStripHit for the left NN
-    if (NNLeftStripEnergy > IonizationEnergy) {
+    // if (NNLeftStripEnergy > IonizationEnergy) {
       MDEEStripHit NNLeftSH = SH;
-      NNLeftSH.m_Energy = NNLeftStripEnergy;
+      NNLeftSH.m_Energy = std::max(NNLeftStripEnergy, 0.0);
```
fhagemann: Note to myself: Looks like this `std::max` to avoid tiny negative energies (e.g. `-1e-12`) and therefore errors in the inverse ecal (`TF1 could not find root in interval [0,16383]`) is not needed if the `InitialChargeCloudSize` is set to something positive instead of `0.0` right now (e.g. `0.01` for 10µm initial charge cloud size)

include/MDEEStripHit.h
ckierans: How about m_TimetoFastShaperPeak to make it very explicit? I agree m_DriftTime might be a little misleading for the non main strips.

src/MSubModuleChargeTransport.cxx
```diff
@@ -136,6 +139,58 @@ bool MSubModuleChargeTransport::Initialize()
...
+
+  } else {
+    return false;
+  }
+
+  // Load depth calibration coefficients
+  DepthCalibration.SetCoeffsFileName(m_DepthCoefficientsFileName);
+  if (DepthCalibration.LoadCoeffsFile(m_DepthCoefficientsFileName) == true) {
+    // Copy depth calibration coefficients
+    m_Coeffs = DepthCalibration.GetCoeffs();
```
ckierans, line 188: Comparing with the previous MSubModuleDepthReadout.cxx, I see you've removed the m_Coeffs_Energy variable. I'm trying to track down what happened to that after PR #152. Are you still reliant on that variable? Is so, where are you calling it?

include/MDEEStripHit.h
ckierans: Or m_FastPeakTime?

src/MSubModuleChargeTransport.cxx
```diff
@@ -262,7 +319,8 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
   MVector Pos = SH.m_SimulatedPositionInDetector;
   double P    = isLV ? Pos.X() : Pos.Y();
   double Q    = isLV ? Pos.Y() : Pos.X();
-  double ΔZ   = isLV ? Pos.Z() + Thickness / 2.0 : Thickness / 2.0 - Pos.Z();
+  double Z    = Pos.Z();
+  double DeltaZ = isLV ? Z + Thickness / 2.0 : Thickness / 2.0 - Z;
```
ckierans, line 324: I prefer the DeltaZ over the symbol that I don't even know how to write. So thanks for bringing that back :)

src/MSubModuleChargeTransport.cxx
```diff
@@ -280,13 +338,35 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
   // TODO: Confirm the correct boundary of the guard ring based on SMEX detector models
   if (ID >= 0 && ID < NStrips && std::abs(Q) <= QWidth/2.0 && std::hypot(P, Q) <= Radius) {
 
+    // Determine the charge drift times in nanoseconds from simulations + stretch/offset from the depth calibration
+    // Set the default to 0ns in case no depth calibration coefficients exist
+    double DriftTime = 1e10;
```
ckierans: Update the comment to reflect the current initialized value

src/MSubModuleChargeTransport.cxx
ckierans: Hmm, good question. If we're sold on moving towards a strip-based inverse depth, then I guess whatever you implement here is temporary. But what about taking the average of some neighboring pixels as a placeholder for the stretch and offset? It's not perfect, but it's an okay guess. Note that once this goes through the forward pipeline, this pixel still won't have a valid depth calibration so it'll result in an error anyways, right?

src/MSubModuleChargeTransport.cxx
ckierans: Good question as to where these gets combined. Looking at our current flow, I'm wondering if it then makes sense to bring MSubModuleDepthReadout immediate after MSubModuleChargeTransport? That allows you to apply the fast threshold and properly combine the m_DriftTimes without keeping the striphit drift times separate throughout all of the energy-focused modules in between? There may be an issues with having the Depth Readout come before the StripTrigger module, but it's not immediately jumping out at me...

This can also be kept as a TODO for now just to get something working for the majority of hits.

src/MSubModuleChargeTransport.cxx
ckierans: Yes, agreed. Leave this simple case for now and add a TODO.

src/MSubModuleChargeTransport.cxx
ckierans: Makes sense, since this is reality.

src/MSubModuleChargeTransport.cxx
ckierans: Interesting idea. It's nice to think about everything being self-consistent. But I wondering if there are any hints that `t = z / v = z / (µ * E)` is not accurate enough? Can the simulations be run without electronics and still be self-consistent with the drift time sims you're using above? Maybe this is a nice-to-have improvement for the future...

src/MSubModuleChargeTransport.cxx
```diff
@@ -307,40 +387,41 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
...
+    // if (NNLeftStripEnergy > IonizationEnergy) {
       MDEEStripHit NNLeftSH = SH;
-      NNLeftSH.m_Energy = NNLeftStripEnergy;
+      NNLeftSH.m_Energy = std::max(NNLeftStripEnergy, 0.0);
+      NNLeftSH.m_DriftTime = DriftTime - 50 * (1 - NNLeftStripEnergy / SH.m_SimulatedEnergy);
       NNLeftSH.m_OppositeStripID = OppositeStripID;
       if (ID > 0) {
         NNLeftSH.m_ROE.SetStripID(ID - 1);
         NNLeftSH.m_IsGuardRing = false;
-        // NNLeftSH.m_IsNearestNeighbor = true;
```
ckierans, line 321: Will this flag then be added in the StripTrigger module instead?

src/MSubModuleDepthReadout.cxx
```diff
@@ -164,211 +132,105 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
-          // Convert drift time to timing by subtracting 4200ns (for now)
-          // TODO: Improve determining the timing from drift times
-          SH.m_Timing = 4200.0 - HoleDriftTime;
-
-          // Smear the timing value based on the given CTD resolution
-          // --> divide by √2 to obtain TAC resolution from CTD resolution
-          if (m_ApplyTimingResolutionCalibration == true) {
+    if (SH.m_IsGuardRing == false) {
+      unsigned int StripID = SH.m_ROE.GetStripID();
+      if (SH.m_DriftTime > -200.0 ) {
```
ckierans: Where does the -200 come from?

src/MSubModuleDepthReadout.cxx
```diff
@@ -164,211 +132,105 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
-        cout << "No Depth Spline for Event with DetID " << DetID << endl;
+        if (SH.m_TAC > 16383) {
+          SH.m_TAC = 16383;
+        }
+      } else {
+        if (g_Verbosity >= c_Info) {
+          cout<<"MSubModuleDepthReadout::AnalyzeEvent: Unphysical drift time."<<endl;
+        }
+        SH.m_TAC = 0;
+        SH.m_HasTriggered = false;
```
ckierans: Shouldn't this be m_HasFastTiming == false instead of m_HasTriggered?

src/MSubModuleDepthReadout.cxx
ckierans: Ah, this answers my question from above. Maybe just add a comment in the code as to why you choose 200 ns as a placeholder.

Why do we need this cut at all? Is this to filter out the unphysical large drift times when we don't have a valid depth calibrations (I thought those would be positive?)?

src/MSubModuleDepthReadout.cxx
```diff
@@ -164,211 +132,105 @@ bool MSubModuleDepthReadout::AnalyzeEvent(MReadOutAssembly* Event)
...
-        } else {
-          if (g_Verbosity >= c_Error) {
-            cout << "ERROR in MSubModuleDepthReadout::LoadSplinesFile: Empty line in depth splines file." << endl;
-          }
-          return false;
+        if (g_Verbosity >= c_Info) {
+          cout<<"MSubModuleDepthReadout::AnalyzeEvent: Unphysical drift time."<<endl;
         }
+        SH.m_TAC = 0;
+        SH.m_HasTriggered = false;
```
ckierans: Same question as above about this flag. I'd need to go back to my notes/git convos, but I thought m_HasTriggered is about the slow circuit?

ckierans (commented):
Sorry this took so long to get through, @fhagemann! Nice rework here. I think things are mostly pretty clear, but I do have a few questions. Nothing major.

include/MDEEStripHit.h
fhagemann: I like `m_FastPeakTime`, because it keeps the name reasonably short while still capturing what it describes. Thanks for this suggestion!

src/MSubModuleChargeTransport.cxx
fhagemann: I am still reliant on that variable but not here in `MSubModuleChargeTransport`, but in `MSubModuleDepthReadout`.

The depth coefficients file carry pixel information on stretch, offset, and timing resolution. `m_CoeffsEnergy` is used to scale the timing resolution to the strip hit energy, but irrelevant for stretch & offset.

In this PR, I'm only moving the stretch & offset part to `MSubModuleChargeTransport`, while keeping the timing resolution (and `m_CoeffsEnergy`) in `MSubModuleDepthReadout` (specifically in line 87):
https://github.com/fhagemann/nuclearizer/blob/bbd325ac35b957c7f6873c845c1c916bf4bdaf60/src/MSubModuleDepthReadout.cxx#L82-L87

src/MSubModuleChargeTransport.cxx
fhagemann: Maybe we can even combine `MSubModuleStripReadout` and `MSubModuleDepthReadout`, and they are doing the same thing for energy and timing (smearing, inverse ecal/taccal, applying thresholds)?

src/MSubModuleChargeTransport.cxx
fhagemann: Brent also brought to my attention that using `Δ` caused some issues compiling this on an older machine.

src/MSubModuleChargeTransport.cxx
fhagemann: Yes, going towards strip-based inverse depth calibration (#158) was motivated exactly by this issue.
Taking an average for the offset might not be the best estimate (we've seen offsets between neighboring pixels jump, especially when the TAC cal was also jumping), but it's probably the best we can do for now + keep a TODO on revisiting this once we tackle #158.

Although the strip hits without depth calibration coefficients might get no valid depth calibration in the forward pipeline, this will also mess up the timing values for the neighboring strips, which have valid depth calibration. Let me think about this a bit more and how to tackle this.

src/MSubModuleChargeTransport.cxx
fhagemann: There are some hints that `t = z / v = z / (µ * E)` might not be accurate enough:
- `E` (electric field) not being homogeneous in the detector due to the non-zero impurity density. The electric field strength increases linearly towards the HV side, which makes this `t` non-linear in `z`
- `µ` (charge carrier mobility) is now set to a constant value, but is actually temperature dependent, dependent on `E` (saturates at higher electric fields), and technically also on the crystal axis orientation.

So, in that case, the `t ~ z` relationship might not necessarily hold.

src/MSubModuleChargeTransport.cxx
fhagemann: This is the plan. We only know after combining strip hits if a strip exceeded the slow threshold or not, and becomes a triggered strip or a "nearest neighbor"

src/MSubModuleDepthReadout.cxx
fhagemann: This is an arbitrary threshold I chose based on the following:
- `m_DriftTime` for the main strip can (in principle) be as tiny as 0ns (or even slightly negative of certain values of the offset determined in the depth calibration)
- nearest-neighbor strips, in this current implementation, can have values that are 50ns smaller
- With this -200, I want to keep all strip hits with "reasonable" timing (NN strips etc., offset values that we might expect), while getting rid of those with "unreasonable" timing

src/MSubModuleDepthReadout.cxx
fhagemann: Good point, I will update this.

src/MSubModuleDepthReadout.cxx
fhagemann: When I played around with other parameterizations for the nearest neighbor timing values, there were some cases in which I was getting very negative values. But you are right: we should also safe-guard against the large positive default value (`1e10`) for strips without depth calibration here.

src/MSubModuleDepthReadout.cxx
fhagemann: Turns out `MDEEStripHit` did not have the flag `m_HasFastTiming`, so I added that now in bb05fb2.

src/MSubModuleChargeTransport.cxx
fhagemann: Plus, we don't know what the slow threshold is, which might only be applied later in `MSubModuleStripReadout`

src/MSubModuleChargeTransport.cxx
```diff
@@ -280,13 +339,35 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
...
+    int PixelCode = 10000*DetID + 100*(isLV ? ID : OppositeStripID) + (isLV ? OppositeStripID : ID);
+
+    // Apply stretch based on Eq. (3) in https://doi.org/10.1016/j.nima.2026.171332
+    // Apply no offset to the electron drift time --> add it fully to the hole (LV) signal
+    auto it = m_Coeffs.find(PixelCode);
+    if (it != m_Coeffs.end()) {
+      const vector<double>& Coeffs = it->second;
+      double Stretch = Coeffs[0];
+      double Offset = isLV ? Coeffs[1] : 0.0;
+      FastPeakTime = (DriftTimeSpline->Eval(Z) + Offset) * Stretch;
```
fhagemann, line 356: Small note: I did not rename `DriftTimeSpline`, because these really are drift times (plus electronics), but then changed everything else to `FastPeakTime`, where then timing effects of charge sharing or early peaking of nearest-neighbor strips will play an effect.

src/MSubModuleChargeTransport.cxx
ckierans: Not a bad idea. Let's keep that as something to explore when you address the proper combination, but leave this as it is for the current PR.

src/MSubModuleChargeTransport.cxx
ckierans: Okay, again, let's think of this as something to work towards, but leave it as is for now.

src/MSubModuleChargeTransport.cxx
```diff
@@ -307,40 +388,38 @@ void MSubModuleChargeTransport::RunChargeTransportForHit(MDEEStripHit& SH, bool
...
-      MDEEStripHit NNRightSH = SH;
-      NNRightSH.m_Energy = NNRightStripEnergy;
-      NNRightSH.m_OppositeStripID = OppositeStripID;
-      if (ID < NStrips - 1) {
-        NNRightSH.m_ROE.SetStripID(ID + 1);
-        NNRightSH.m_IsGuardRing = false;
-        // NNRightSH.m_IsNearestNeighbor = true;
-      } else {
-        NNRightSH.m_ROE.SetStripID(NStrips);
-        NNRightSH.m_IsGuardRing = true;
```
parshadkp, line 340: Sorry, this is super late but does this assign a GR hit even if there was no energy transported to the GR? Is this okay for a GR veto?

parshadkp (commented):
Hi @fhagemann, sorry this is super late but I was looking through the changes and found this one potential issue. Let me know what you think.

src/MSubModuleChargeTransport.cxx
fhagemann: Yes, it does, and it might trigger a GR veto in its current version.
But, by applying GR energy thresholds later in `MSubModuleStripTrigger` or `MSubModuleStripReadout` (I think this is not implemented right now), these events would then be filtered out.

parshadkp: That makes sense. Once we have the thresholds implemented then I can check if the number of GR vetoes goes down.

Issue #105, opened 2026-03-13
Data base for time dependent calibration file handling


zoglauer:
We have to design a data storage system for our calibration files. 

Key challenge is that the calibration and operation parameter of COSI change with time, e.g. from T1 -> T2 use thresholds A, from T2 -> T3 use thresholds B.

CERN experiments use data bases:
https://s3.cern.ch/inspire-prod-files-0/0610468e7d178025cc735db3ca7054d3

We should think about what we want to do, and then implement is

Issue #104, opened 2026-03-10
Read bias voltages and temperatures of the detectors from a database for the DEE


fhagemann:
Right now, I am using hard-coded values in the DEE.

PR #171, opened 2026-06-29
CHG: MStripHit cleanup

zoglauer:
I did the same cleanup for MStripHit.

But here are a few more old member functions around.

I would like to remove:
  //! DEPRECATED: Set whether the strip is on the low voltage side using IsLowVoltageStrip instead
  void IsXStrip(bool LowVoltageSide) { m_ReadOutElement->IsLowVoltageStrip(LowVoltageSide); }
  //! DEPRECATED: Return whether the strip is on the low voltage side using IsLowVoltageStrip instead
  bool IsXStrip() const { return m_ReadOutElement->IsLowVoltageStrip(); }
--> That is now IsLowVoltageStrip(..) REMOVE?

  //! Set the timing in nanoseconds
  void SetTiming(double Timing) { m_Timing = Timing; }
  //! Return the timing in nanoseconds
  double GetTiming() const { return m_Timing; }

  //! Set the timing resolution
  void SetTimingResolution(double TimingResolution) { m_TimingResolution = TimingResolution; }
  //! Return the timing resolution
  double GetTimingResolution() const { return m_TimingResolution; }
--> Has been replaced by the same function with name TAC instead of timing. REMOVE?

  //! Set the temperature of the relevant preamp (in degrees C)
  void SetPreampTemp(double PreampTemp) { m_PreampTemp = PreampTemp; }
  //! Return the temperature of the relevant preamp (in degrees C)
  double GetPreampTemp() const { return m_PreampTemp; }
--> What about that - we definitely have no preamps, but maybe some temperature dependence? But do we plan to store the temperature here?

### DIFF
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -83,7 +83,7 @@ class MStripHit
   //! TODO: Switch to unsigned int: Issue #176
   //! Set the measured ADC units of the strip
   void SetADCUnits(double ADCUnits) { m_ADCUnits = ADCUnits; }
-  //! Return the measured the ADC units of the strip
+  //! Return the measured ADC units of the strip
   double GetADCUnits() const { return m_ADCUnits; }
 
   //! Set the calibrated energy (= calibrated ADC's) in keV
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -107,7 +107,7 @@ class MStripHit
 
   //! Set the timing (= calibrated TAC) in nanoseconds
   void SetTiming(double Timing) { m_Timing = Timing; }
-  //! Return the timing (= calibrated TAC)in nanoseconds
+  //! Return the timing (= calibrated TAC) in nanoseconds
   double GetTiming() const { return m_Timing; }
 
   //! Set the timing resolution
@@ -139,7 +139,7 @@ class MStripHit
   //! Return whether the strip has triggered (ADC values above slow threshold)
   bool HasTriggered() const { return m_HasTriggered; }
 
-  //! REVIEW: Should this be just HasTiming() - timing is calibrated TAC?
+  //! TODO: Rename to HasTiming()
   //! Set the calibrated-timing flag
   void HasCalibratedTiming(bool CalibratedTiming) { m_HasCalibratedTiming = CalibratedTiming; }
   //! Return whether the strip timing has been calibrated
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -40,15 +40,18 @@ class MStripHit
  public:
   //! Default constructor
   MStripHit();
+  //! Copy constructor is disabled because this class owns a m_ReadOutElement pointer
+  MStripHit(const MStripHit&) = delete;
+  //! Assignment is disabled because this class owns a m_ReadOutElement pointer
+  MStripHit& operator=(const MStripHit&) = delete;
   //! Default destructor
   virtual ~MStripHit();
-  //! Copy constructor is disabled because this class owns m_ReadOutElement
-  MStripHit(const MStripHit&) = delete;
-  //! Assignment is disabled because this class owns m_ReadOutElement
-  MStripHit& operator=(const MStripHit&) = delete;
 
   //! Reset all data
   void Clear();
+
+
+  // The read-out element
 
   //! Return the read-out element
   //! Ownership stays with this object
@@ -64,46 +67,52 @@ class MStripHit
   //! Return the strip ID
   unsigned int GetStripID() const { return m_ReadOutElement->GetStripID(); }
 
-  //! DEPRECATED: Set whether the strip is on the low voltage side using IsLowVoltageStrip instead
+  //! DEPRECATED: Set whether the strip is on the low-voltage side using IsLowVoltageStrip instead
   void IsXStrip(bool LowVoltageSide) { m_ReadOutElement->IsLowVoltageStrip(LowVoltageSide); }
-  //! DEPRECATED: Return whether the strip is on the low voltage side using IsLowVoltageStrip instead
+  //! DEPRECATED: Return whether the strip is on the low-voltage side using IsLowVoltageStrip instead
   bool IsXStrip() const { return m_ReadOutElement->IsLowVoltageStrip(); }
 
-  //! Set whether the strip is on the low voltage side
+  //! Set whether the strip is on the low-voltage side
   void IsLowVoltageStrip(bool LowVoltageSide) { m_ReadOutElement->IsLowVoltageStrip(LowVoltageSide); }
-  //! Return whether the strip is on the low voltage side
+  //! Return whether the strip is on the low-voltage side
   bool IsLowVoltageStrip() const { return m_ReadOutElement->IsLowVoltageStrip(); }
 
-  //! Set whether the strip has triggered
-  void HasTriggered(bool HasTriggered) { m_HasTriggered = HasTriggered; }
-  //! Return whether the strip has triggered
-  bool HasTriggered() const { return m_HasTriggered; }
 
+  // Energy section:
+
+  //! REVIEW: Why is this not unsigned int - that is what we measure?
+  //! Set the measured ADC units of the strip
+  void SetADCUnits(double ADCUnits) { m_ADCUnits = ADCUnits; }
+  //! Return measured the ADC units of the strip
+  double GetADCUnits() const { return m_ADCUnits; }
+
+  //! REVIEW: What are we using those for - they are not used anywhere - I would always correct the energy not the raw data
+  //!           We might use it for temperature correction - or not
   //! Set the uncorrected ADC units of the strip (before common-mode correction)
   void SetUncorrectedADCUnits(double UncorrectedADCUnits) { m_UncorrectedADCUnits = UncorrectedADCUnits; }
   //! Return the uncorrected ADC units of the strip (before common-mode correction)
   double GetUncorrectedADCUnits() const { return m_UncorrectedADCUnits; }
-
-  //! Set the ADC units of the strip
-  void SetADCUnits(double ADCUnits) { m_ADCUnits = ADCUnits; }
-  //! Return the ADC units of the strip
-  double GetADCUnits() const { return m_ADCUnits; }
 
   //! Set the calibrated energy
   void SetEnergy(double Energy) { m_Energy = Energy; }
   //! Return the calibrated energy
   double GetEnergy() const { return m_Energy; }
 
-  //! Set the energy resolution (sigma)
+  //! Set the energy resolution (1 sigma)
   void SetEnergyResolution(double EnergyResolution) { m_EnergyResolution = EnergyResolution; }
-  //! Return the energy resolution
+  //! Return the energy resolution (1 sigma)
   double GetEnergyResolution() const { return m_EnergyResolution; }
 
-  //! Set the TAC
+
+  // Timing:
+
+  //! REVIEW: Why is TAC double, when we measure unsigned ints?
+  //! Set the measured TAC value of the strip (arrival timing)
   void SetTAC(double TAC) { m_TAC = TAC; }
-  //! Return the TAC
+  //! Return the measured TAC value of the strip (arrival timing)
   double GetTAC() const { return m_TAC; }
 
+  //! REVIEW: This is not used - timing should have a resolution
   //! Set the TAC resolution
   void SetTACResolution(double TACResolution) { m_TACResolution = TACResolution; }
   //! Return the TAC resolution
@@ -119,23 +128,26 @@ class MStripHit
   //! Return the timing resolution
   double GetTimingResolution() const { return m_TimingResolution; }
 
+
+  // Temperature:
+
+  //! REVIEW: Hold-over from balloon and balloon temperature calibration
   //! Set the temperature of the relevant preamp (in degrees C)
   void SetPreampTemp(double PreampTemp) { m_PreampTemp = PreampTemp; }
   //! Return the temperature of the relevant preamp (in degrees C)
   double GetPreampTemp() const { return m_PreampTemp; }
 
-  //! Add origins from the simulation and remove duplicates
-  void AddOrigins(const vector<int>& Origins);
-  //! Return the origins from the simulation
-  vector<int> GetOrigins() const { return m_Origins; }
+
+  // Flags:
 
   //! Set the guard ring flag
   void IsGuardRing(bool GuardRing) { m_IsGuardRing = GuardRing; }
-  //! Return whether the strip is a guard ring
+  //! Return whether the strip is a guard-ring hit
   bool IsGuardRing() const { return m_IsGuardRing; }
+
   //! Set the nearest-neighbor flag
   void IsNearestNeighbor(bool NearestNeighbor) { m_IsNearestNeighbor = NearestNeighbor; }
-  //! Return whether the strip is a nearest neighbor
+  //! Return whether the strip is a nearest-neighbor hit
   bool IsNearestNeighbor() const { return m_IsNearestNeighbor; }
 
   //! Set the fast-timing flag
@@ -143,22 +155,40 @@ class MStripHit
   //! Return whether the strip timing is fast
   bool HasFastTiming() const { return m_HasFastTiming; }
 
+  //! REVIEW: What kind of trigger is this: above slow or above fast?
+  //! Set whether the strip has triggered
+  void HasTriggered(bool HasTriggered) { m_HasTriggered = HasTriggered; }
+  //! Return whether the strip has triggered
+  bool HasTriggered() const { return m_HasTriggered; }
+
+  //! REVIEW: Should this be just HasTiming() - timing is calibrated TAC?
   //! Set the calibrated-timing flag
   void HasCalibratedTiming(bool CalibratedTiming) { m_HasCalibratedTiming = CalibratedTiming; }
   //! Return whether the strip timing has been calibrated
   bool HasCalibratedTiming() const { return m_HasCalibratedTiming; }
+
+
+  // Parsing:
 
   //! Return the bitwise strip-hit flags
   unsigned int MakeFlags();
   //! Update the strip-hit flags from a bit mask
   void ParseFlags(unsigned int Flags);
 
-  //! Parse a strip hit from a line
+  //! Parse a strip hit from a line in DAT format
   bool Parse(const MString& Line, int Version = 1);
-  //! Write the strip hit to a file stream
+  //! Stream the strip hit in Nuclearizer's DAT format
   bool StreamDat(ostream& S, int Version = 1);
   //! Stream the strip hit in MEGAlib's ROA format
   void StreamRoa(ostream& S, bool WithADC = true, bool WithTAC = true, bool WithEnergy = false, bool WithTiming = false, bool WithTemperature = false, bool WithFlags = false, bool WithOrigins = false);
+
+
+  // Simulation:
+
+  //! Add origins from the simulation and remove duplicates
+  void AddOrigins(const vector<int>& Origins);
+  //! Return the origins from the simulation
+  vector<int> GetOrigins() const { return m_Origins; }
 
 
   // protected methods:
@@ -177,8 +207,7 @@ class MStripHit
  private:
   //! Read-out element
   MReadOutElementDoubleStrip* m_ReadOutElement;
-  //! True if the strip has triggered
-  bool m_HasTriggered;
+
   //! ADC units before all corrections
   double m_UncorrectedADCUnits;
   //! ADC units after any correction
@@ -187,7 +216,8 @@ class MStripHit
   double m_Energy;
   //! Energy resolution
   double m_EnergyResolution;
-  //! TAC timing
+
+  //! The measured TAC timing values
   double m_TAC;
   //! TAC timing resolution
   double m_TACResolution;
@@ -195,14 +225,16 @@ class MStripHit
   double m_Timing;
   //! Timing resolution in nanoseconds
   double m_TimingResolution;
+
   //! Temperature of the preamp
   double m_PreampTemp;
 
-  //! True if the hit is a guard ring
+  //! True if the hit is a guard ring hit
   bool m_IsGuardRing;
-  //! True if the hit is a nearest neighbor
+  //! True if the hit is a nearest neighbor hit
   bool m_IsNearestNeighbor;
-
+  //! True if the strip has triggered
+  bool m_HasTriggered;
   //! True if the hit has fast timing
   bool m_HasFastTiming;
   //! True if the hit has calibrated timing
@@ -210,6 +242,7 @@ class MStripHit
 
   //! Origin interaction IDs from the simulation
   vector<int> m_Origins;
+
 
 #ifdef ___CLING___
  public:
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -86,9 +86,9 @@ class MStripHit
   //! Return measured the ADC units of the strip
   double GetADCUnits() const { return m_ADCUnits; }
 
-  //! Set the calibrated energy
+  //! Set the calibrated energy (= calibrated ADC's) in keV
   void SetEnergy(double Energy) { m_Energy = Energy; }
-  //! Return the calibrated energy
+  //! Return the calibrated energy (= calibrated ADC's) in keV
   double GetEnergy() const { return m_Energy; }
 
   //! Set the energy resolution (1 sigma)
@@ -105,9 +105,9 @@ class MStripHit
   //! Return the measured TAC value of the strip (arrival timing)
   double GetTAC() const { return m_TAC; }
 
-  //! Set the timing in nanoseconds
+  //! Set the timing (= calibrated TAC) in nanoseconds
   void SetTiming(double Timing) { m_Timing = Timing; }
-  //! Return the timing in nanoseconds
+  //! Return the timing (= calibrated TAC)in nanoseconds
   double GetTiming() const { return m_Timing; }
 
   //! Set the timing resolution
@@ -186,16 +186,16 @@ class MStripHit
   //! Read-out element
   MReadOutElementDoubleStrip* m_ReadOutElement;
 
-  //! ADC units after any correction
+  //! Measured ADC units
   double m_ADCUnits;
-  //! Calibrated energy
+  //! Calibrated energy (= calibrated ADC's) in keV
   double m_Energy;
-  //! Energy resolution
+  //! Energy resolution in keV (1 sigma)
   double m_EnergyResolution;
 
   //! The measured TAC timing values
   double m_TAC;
-  //! Timing in nanoseconds
+  //! Timing (= calibrated TAC) in nanoseconds
   double m_Timing;
   //! Timing resolution in nanoseconds
   double m_TimingResolution;
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -67,9 +67,9 @@ class MStripHit
   //! Return the strip ID
   unsigned int GetStripID() const { return m_ReadOutElement->GetStripID(); }
 
-  //! DEPRECATED: Set whether the strip is on the low-voltage side using IsLowVoltageStrip instead
+  //! DEPRECATED: Remove, but we need to remove it from remaining classes that use it first: Issue #172
   void IsXStrip(bool LowVoltageSide) { m_ReadOutElement->IsLowVoltageStrip(LowVoltageSide); }
-  //! DEPRECATED: Return whether the strip is on the low-voltage side using IsLowVoltageStrip instead
+  //! DEPRECATED: Remove, but we need to remove it from remaining classes that use it first: Issue #172
   bool IsXStrip() const { return m_ReadOutElement->IsLowVoltageStrip(); }
 
   //! Set whether the strip is on the low-voltage side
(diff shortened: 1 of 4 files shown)


fhagemann:
I'm not sure that `TAC` and `Timing` mean the same thing (also I differentiate between the two in the DEE).
Based on the short description in #11 where `m_TAC` was introduced, I believe `m_TAC` is the **raw** TAC value in ADC units and `Timing` is the **calibated** TAC value in nanoseconds.
While I agree, that we might not need both `TACResolution` and `TimingResolution`, I would still vote for keeping both `m_Timing` and `m_TAC`.

fhagemann:
Also, when I do data/sim comparisons, I usually write both `TAC` (raw TAC value) and `Timing` (calibrated TAC value) to an ROA file, and they have different values.

The only problem here is that we are inconsistent with types between nuclearizer and MEGAlib:
`m_Timing` here in nuclearizer is of type `double` (which for nanoseconds makes sense I would say), whereas in MEGAlib `Timing` is defined as `unsigned int` -- more consistent with this actually being an ADC value which it isn't in nuclearizer anymore (https://github.com/zoglauer/megalib/blob/develop-cosi/src/fretalon/base/inc/MReadOutDataTiming.h). This requires me to change `Timing` in MEGAlib on my laptop from `unsigned int` to  `double` to read/write ROA files.

fhagemann:
@zoglauer related to this, I would like to resurface the discussion we had back in https://github.com/cositools/nuclearizer/pull/120, that addressed exactly this inconsistency between `TAC` and `Timing`.

include/MStripHit.h
```diff
@@ -41,46 +42,51 @@ class MStripHit
...
+  //! Return the strip ID
   unsigned int GetStripID() const { return m_ReadOutElement->GetStripID(); }
 
-  //! Set the strip type (x/y)
-  void IsXStrip(bool PositiveStrip) { m_ReadOutElement->IsLowVoltageStrip(PositiveStrip); }
-  //! Return the strip type (x/y)
+  //! DEPRECATED: Set whether the strip is on the low voltage side using IsLowVoltageStrip instead
+  void IsXStrip(bool LowVoltageSide) { m_ReadOutElement->IsLowVoltageStrip(LowVoltageSide); }
+  //! DEPRECATED: Return whether the strip is on the low voltage side using IsLowVoltageStrip instead
   bool IsXStrip() const { return m_ReadOutElement->IsLowVoltageStrip(); }
```
fhagemann: This can go, as long as it's not used anywhere else in nuclearizer (I see some occurrences in the old DEE files, in `apps/ThresholdDeterminator.cxx`, and in `src/MModuleLoaderMeasurementsROA.cxx` -- the latter would need to be updated then).

include/MStripHit.h
```diff
@@ -41,46 +42,51 @@ class MStripHit
...
-  //! Return the uncorrected ADCUnits of the strip (before common-mode correction)
+  //! Return the uncorrected ADC units of the strip (before common-mode correction)
   double GetUncorrectedADCUnits() const { return m_UncorrectedADCUnits; }
 
-  //! Set the ADCUnits of the strip
+  //! Set the ADC units of the strip
   void SetADCUnits(double ADCUnits) { m_ADCUnits = ADCUnits; }
-  //! Return the ADCUnits of the strip
+  //! Return the ADC units of the strip
   double GetADCUnits() const { return m_ADCUnits; }
```
fhagemann: Just from looking at this, I wouldn't know what this is (energy or timing in ADC units, before some common-mode correction?)

include/MStripHit.h
```diff
@@ -41,46 +42,51 @@ class MStripHit
...
-  //! Return the strip type (positive or negative)
+  //! Set whether the strip is on the low voltage side
+  void IsLowVoltageStrip(bool LowVoltageSide) { m_ReadOutElement->IsLowVoltageStrip(LowVoltageSide); }
+  //! Return whether the strip is on the low voltage side
   bool IsLowVoltageStrip() const { return m_ReadOutElement->IsLowVoltageStrip(); }
 
   //! Set whether the strip has triggered
   void HasTriggered(bool HasTriggered) { m_HasTriggered = HasTriggered; }
   //! Return whether the strip has triggered
   bool HasTriggered() const { return m_HasTriggered; }
```
fhagemann: "triggered" as in above slow threshold (energy triggered), or could this also be exceeding the fast threshold (timing), without exceeding the slow threshold, what we see for some peak-dip pulse shapes in non-charge-collecting neighbor strips?

I believe its the former, because I also see `HasFastTiming`, but maybe make this explicit in the comment.

include/MStripHit.h
```diff
@@ -103,58 +109,58 @@ class MStripHit
...
-  //! Return the Timing resolution
+  //! Return the timing resolution
   double GetTimingResolution() const { return m_TimingResolution; }
 
-  //! Set the Temperature of the relavent preamp (in degrees C)
+  //! Set the temperature of the relevant preamp (in degrees C)
   void SetPreampTemp(double PreampTemp) { m_PreampTemp = PreampTemp; }
-  //! Return the Temperature of the relavent preamp (in degrees C)
+  //! Return the temperature of the relevant preamp (in degrees C)
   double GetPreampTemp() const { return m_PreampTemp; }
```
fhagemann: This only seems to be used in the old DEE files, and we have an issue to keep track on how to deal with in-flight operational parameters, like temperature of the ASICs (#104 / #105)

include/MStripHit.h
```diff
@@ -103,58 +109,58 @@ class MStripHit
...
-  //! Return a boolean indicating whether the strip timing is fast;
+  //! Return whether the strip timing is fast
   bool HasFastTiming() const { return m_HasFastTiming; }
 
-  //! Set the Calibrated Timing flag
+  //! Set the calibrated-timing flag
   void HasCalibratedTiming(bool CalibratedTiming) { m_HasCalibratedTiming = CalibratedTiming; }
-  //! Return a boolean indicating whether the strip timing has been calibrated;
+  //! Return whether the strip timing has been calibrated
   bool HasCalibratedTiming() const { return m_HasCalibratedTiming; }
```
fhagemann, line 146: Is this because `m_Timing` used to first hold uncalibrated TAC and then calibrated TAC. Or is this to flag all strips where we don't have a TAC cal available for example (guard ring for example)?

include/MStripHit.h
```diff
@@ -169,44 +175,45 @@ class MStripHit
...
   //! TAC timing
   double m_TAC;
   //! TAC timing resolution
   double m_TACResolution;
-  //! Timing in ns
+  //! Timing in nanoseconds
   double m_Timing;
-  //! Timing resolution in ns
+  //! Timing resolution in nanoseconds
   double m_TimingResolution;
```
fhagemann: My understanding here would be:
```suggestion
  //! Uncalibrated TAC in ADC units
  double m_TAC;
  //! Resolution of uncalibrated TAC in ADC units
  double m_TACResolution; // <-- this can probably go, makes no physical sense
  //! Calibrated TAC in nanoseconds
  double m_Timing;
  //! Resolution of calibrated TAC in nanoseconds
  double m_TimingResolution;
```

include/MStripHit.h
```diff
@@ -169,44 +175,45 @@ class MStripHit
...
-  //! Temperature of Preamp
+  //! Temperature of the preamp
   double m_PreampTemp;
```
fhagemann: This can go --> #104 / #105

include/MStripHit.h
```diff
@@ -169,44 +175,45 @@ class MStripHit
...
 
-  //! Flags denoting the type of strip hit
+  //! True if the hit is a guard ring
```
fhagemann:
```suggestion
  //! True if the hit is a guard ring hit
```

include/MStripHit.h
```diff
@@ -169,44 +175,45 @@ class MStripHit
...
   bool m_IsGuardRing;
+  //! True if the hit is a nearest neighbor
```
fhagemann:
```suggestion
  //! True if the hit is a nearest neighbor hit
```

src/MStripHit.cxx
fhagemann, line 205: This is what I was referring to: we can write both (uncalibrated) `m_TAC` and (calibrated) `m_Timing` to ROA files, but currently we cannot READ the `Timing` because MEGAlib parses it as `unsigned int` where here in nuclearizer we define it as `double`.

src/MStripHit.cxx
```diff
@@ -229,12 +236,12 @@ void MStripHit::StreamRoa(ostream& S, bool WithADC, bool WithTAC, bool WithEnerg
...
-  //! Currently, 3 bits:
-  //!   v = Has fast timing
-  //!    v = Is a nearest neighbor
-  //!     v = Is a guard ring
-  //! 0b111u
+  // Return the bitwise strip-hit flags
+  // Currently, 3 bits:
+  //   v = Has fast timing
+  //    v = Is a nearest neighbor
+  //     v = Is a guard ring
```
fhagemann:
```suggestion
  //    v = Is a nearest neighbor hit
  //     v = Is a guard ring hit
```

src/MStripHit.cxx
```diff
@@ -256,14 +263,14 @@ unsigned int MStripHit::MakeFlags()
...
-  //!    v = Is a nearest neighbor
-  //!     v = Is a guard ring
-  //! 0b111u
-
-  // "Flags & 0b001u" extracts bit 0, "!= 0u" turns it into an explicit bool.
+  // Update the strip-hit flags from a bit mask
+  // Currently, 3 bits:
+  //   v = Has fast timing
+  //    v = Is a nearest neighbor
+  //     v = Is a guard ring
```
fhagemann, line 260:
```suggestion
  //    v = Is a nearest neighbor hit
  //     v = Is a guard ring hit
```

unittests/UTNStripHit.cxx
```diff
@@ -245,35 +245,35 @@ bool UTNStripHit::TestMakeParseFlags()
...
   H.IsNearestNeighbor(false);
   H.HasFastTiming(true);
   Passed = Evaluate("MakeFlags()", "fast timing only", "MakeFlags() returns 4 when only HasFastTiming is set",
                     H.MakeFlags(), (unsigned int) 0b100) && Passed;
 
-  // All three flags → 7
+  // All three flags -> 7
   H.IsGuardRing(true);
   H.IsNearestNeighbor(true);
   H.HasFastTiming(true);
```
fhagemann, line 264: For the unit tests, this is fine, but I would argue that we never flag a guard ring hit also as nearest neighbor 😉

zoglauer:
The TAC / timing issues is that for the balloon "timing" was TAC and thus unsigned int. And now we reused is as calibrated TAC...
OK. I changed it in MEAGlib develop-cosi to double, so that should be in alignment now.

fhagemann:
One question on your change to MEGAlib develop-cosi:
https://github.com/zoglauer/megalib/commit/0c2864c6a1411ce9e5c054b3e664123a5adc716f#r190660654

The `MReadOutDataTiming::Parse` line reading `m_Timing` is commented out. Why is that?

include/MStripHit.h
zoglauer: I made an issue since a few other parts in the code still use it.

zoglauer:
I did some updates, but we will go through them tomorrow during the calibration meeting

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
-  bool HasTriggered() const { return m_HasTriggered; }
 
-  //! Set the uncorrected ADCUnits of the strip (before common-mode correction)
-  void SetUncorrectedADCUnits(double UncorrectedADCUnits) { m_UncorrectedADCUnits = UncorrectedADCUnits; }
-  //! Return the uncorrected ADCUnits of the strip (before common-mode correction)
-  double GetUncorrectedADCUnits() const { return m_UncorrectedADCUnits; }
+  // Energy section:
 
-  //! Set the ADCUnits of the strip
+  //! REVIEW: Why is this not unsigned int - that is what we measure?
```
ckierans: Yes, this is what we measure, so unsigned int makes sense. 

But is there a reason why we wouldn't want this as unsigned int for the DEE @fhagemann ? I assume we probably don't care about the small loss of accuracy from double -> int if we rounded the ADC value that is calculated in the DEE.

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
-  //! Return the ADCUnits of the strip
+  //! Return measured the ADC units of the strip
   double GetADCUnits() const { return m_ADCUnits; }
 
+  //! REVIEW: What are we using those for - they are not used anywhere - I would always correct the energy not the raw data
+  //!           We might use it for temperature correction - or not
+  //! Set the uncorrected ADC units of the strip (before common-mode correction)
+  void SetUncorrectedADCUnits(double UncorrectedADCUnits) { m_UncorrectedADCUnits = UncorrectedADCUnits; }
+  //! Return the uncorrected ADC units of the strip (before common-mode correction)
+  double GetUncorrectedADCUnits() const { return m_UncorrectedADCUnits; }
```
ckierans: Looks like a hold-over from a previous time when common-mode corrections were used. But this isn't something we currently do. 

I would also argue for correcting the energy not the ADC, so perhaps we can remove these for now and then implement the temperature correction on the energy at a later time.

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
+  // Timing:
+
+  //! REVIEW: Why is TAC double, when we measure unsigned ints?
+  //! Set the measured TAC value of the strip (arrival timing)
   void SetTAC(double TAC) { m_TAC = TAC; }
-  //! Return the TAC
+  //! Return the measured TAC value of the strip (arrival timing)
   double GetTAC() const { return m_TAC; }
 
+  //! REVIEW: This is not used - timing should have a resolution
```
ckierans: Yes, I agree we can remove these

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
   void SetEnergyResolution(double EnergyResolution) { m_EnergyResolution = EnergyResolution; }
-  //! Return the calibrated energy
+  //! Return the energy resolution (1 sigma)
   double GetEnergyResolution() const { return m_EnergyResolution; }
 
-  //! Set the TAC
+
+  // Timing:
+
+  //! REVIEW: Why is TAC double, when we measure unsigned ints?
```
ckierans: Same as above for ADC. We're measuring ints, so unless we need double for the DEE (which I don't think is the case) then we should change to int.

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
   void SetTimingResolution(double TimingResolution) { m_TimingResolution = TimingResolution; }
-  //! Return the Timing resolution
+  //! Return the timing resolution
   double GetTimingResolution() const { return m_TimingResolution; }
 
-  //! Set the Temperature of the relavent preamp (in degrees C)
+
+  // Temperature:
+
+  //! REVIEW: Hold-over from balloon and balloon temperature calibration
```
ckierans: Yes, these are old and we'll likely handle the temperature in a different way. I think you can remove these.

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
-  //! Return a boolean indicating whether the strip is a Nearest Neighbor
+  //! Return whether the strip is a nearest-neighbor hit
   bool IsNearestNeighbor() const { return m_IsNearestNeighbor; }
-    
-  //! Set the Fast Timing flag
+
+  //! Set the fast-timing flag
   void HasFastTiming(bool FastTiming) { m_HasFastTiming = FastTiming; }
-  //! Return a boolean indicating whether the strip timing is fast;
+  //! Return whether the strip timing is fast
```
ckierans: "Return whether the strip has passed the fast threshold"

include/MStripHit.h
```diff
@@ -33,128 +33,164 @@
...
   bool HasFastTiming() const { return m_HasFastTiming; }
 
-  //! Set the Calibrated Timing flag
+  //! REVIEW: What kind of trigger is this: above slow or above fast?
```
ckierans: I assume this is the slow threshold since the above fast-timing flag indicates the fast threshold. I don't know if there is a good use case for this flag in the forward pipeline, and it doesn't look like it's currently used.

include/MStripHit.h
ckierans: I assume this is a hold-over, as you say, but @NicoleRodriguezCavero is using this in the TAC Calibration module. I think we should be using the guard ring and fast threshold flags to determine when we do not have a valid TAC cal, but Nicole can better comment on her intention with this variable.

include/MStripHit.h
```diff
@@ -169,44 +205,48 @@ class MStripHit
...
+  //! Read-out element
   MReadOutElementDoubleStrip* m_ReadOutElement;
-  //! Strp has triggered
-  bool m_HasTriggered;
-  //! ADCUnits before all corrections
+
+  //! ADC units before all corrections
   double m_UncorrectedADCUnits;
-  //! ADCUnits after any correction
+  //! ADC units after any correction
```
ckierans: I think we should just have a single ADC here and remove the Uncorrected version of the variable.

include/MStripHit.h
```diff
@@ -169,44 +205,48 @@ class MStripHit
...
   double m_Energy;
-  //! The energy resolution
+  //! Energy resolution
   double m_EnergyResolution;
-  //! TAC timing
+
+  //! The measured TAC timing values
   double m_TAC;
   //! TAC timing resolution
   double m_TACResolution;
```
ckierans: See comment above about this variable likely not being needed

include/MStripHit.h
```diff
@@ -169,44 +205,48 @@ class MStripHit
...
+  //! True if the hit is a nearest neighbor hit
   bool m_IsNearestNeighbor;
-
-  //! Flag indicating whether the hit has fast timing
+  //! True if the strip has triggered
+  bool m_HasTriggered;
+  //! True if the hit has fast timing
   bool m_HasFastTiming;
-  //! Flag indicating whether the hit has calibrated timing
+  //! True if the hit has calibrated timing
```
ckierans, line 212: Not sure this is needed.

include/MStripHit.h
fhagemann: I mean, if we want to mimic data in the DEE, we want to convert the „uncalibrated“ energy in ADC units to an unsigned int to account for the loss in precision due to the discretization that we have intrinsically in data. 

From the DEE side, changing this to unsigned int should not be an issue

zoglauer:
OK. I am done for the moment.
I will switch from HasCalibratedTiming, to HasTiming in the next PR, since this would change to many unrelated file for this cleanup PR.

zoglauer:
Final, final changes

src/MReadOutAssembly.cxx
```diff
@@ -647,6 +647,8 @@ void MReadOutAssembly::StreamEvta(ostream& S)
 void MReadOutAssembly::StreamRoa(ostream& S, bool WithADCs, bool WithTACs, bool WithEnergies, bool WithTimings, bool WithTemperatures, bool WithFlags, bool WithOrigins, bool WithNearestNeighbors)
 {
   // Stream the read-out assembly in MEGAlib's ROA format
+  //
+  // WithTemperatures is currently not used, since we don't have that housekeeping info at the moment
```
fhagemann, line 651: The ROA event saver still allows to set `Include temperatures`, so this is just going to get ignored, but we keep the button in the GUI?

Image

src/MStripHit.cxx
```diff
@@ -116,15 +113,15 @@ bool MStripHit::Parse(const MString& Line, int Version)
...
     unsigned int flags;
-    int N = sscanf(&line[3], "%u %c %u %d %lf %lf %lf %lf %lf %u",
+    int N = sscanf(&line[3], "%u %c %u %d %lf %lf %lf %lf %u",
                    &det_id, &pos_strip, &strip_id, &has_triggered,
-                   &timing, &un_adc, &adc, &energy, &energy_res, &flags);
-    if (N != 10) {
-      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: malformed SH line with "<<N<<" fields instead of 10"<<endl;
+                   &timing, &adc, &energy, &energy_res, &flags);
+    if (N != 9) {
+      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: malformed SH line with "<<N<<" fields instead of 9"<<endl;
```
fhagemann, line 124: This means all previously generated files would need to be re-generated?

include/MStripHit.h
fhagemann: I still don't know how this is relates to `HasFastTiming`.
The comment `//! Return whether the strip timing has been calibrated` suggests that this flag indicates that `m_Timing` was calculated from `m_TAC`, such that we can trust whatever is stored in `m_Timing` (?)

include/MStripHit.h
fhagemann: --> only reason I see would be to indicate than `m_Timing` doesn't store a trash value (?), but then we should check if this flag is actually set accordingly.

src/MStripHit.cxx
fhagemann: Please resolve this comment if you are not planning on incorporating it

src/MStripHit.cxx
fhagemann: Please resolve this comment if you are not planning on incorporating it

include/MStripHit.h
```diff
@@ -33,128 +33,142 @@
...
-  //! Return the uncorrected ADCUnits of the strip (before common-mode correction)
-  double GetUncorrectedADCUnits() const { return m_UncorrectedADCUnits; }
+  // Energy section:
 
-  //! Set the ADCUnits of the strip
+  //! TODO: Switch to unsigned int: Issue #176
+  //! Set the measured ADC units of the strip
   void SetADCUnits(double ADCUnits) { m_ADCUnits = ADCUnits; }
-  //! Return the ADCUnits of the strip
+  //! Return measured the ADC units of the strip
```
ckierans: `//! Return the measured ADC units of the strip`

include/MStripHit.h
```diff
@@ -33,128 +33,142 @@
...
   void HasFastTiming(bool FastTiming) { m_HasFastTiming = FastTiming; }
-  //! Return a boolean indicating whether the strip timing is fast;
+  //! Return whether the strip has passed the fast threshold
   bool HasFastTiming() const { return m_HasFastTiming; }
 
-  //! Set the Calibrated Timing flag
+  //! Set whether the strip has triggered (ADC values above slow threshold)
+  void HasTriggered(bool HasTriggered) { m_HasTriggered = HasTriggered; }
+  //! Return whether the strip has triggered (ADC values above slow threshold)
+  bool HasTriggered() const { return m_HasTriggered; }
```
ckierans, line 140: I don't understand the use-case of this flag. Isn't this redundant with the IsNearestNeightbor flag? Is there ever a time the HasTriggered will be used when IsNearestNeighbor is not sufficient?

include/MStripHit.h
ckierans: I'm not sure we need this flag. Nicole sets this in her TAC cal module, but no other module references it, I don't think. 

I agree with your interpretation of what the comment implies, Felix, but I'm not sure it makes sense to keep this around since m_Timing would be 0, as initiated, unless the strip timing has been calibrated, right?

src/MStripHit.cxx
ckierans: Shouldn't this be "nearest neighbor strip hit" and "guard ring strip hit" vs just "hit"

src/MStripHit.cxx
ckierans: Same as above. Use "strip hit" not "hit"

src/MStripHit.cxx
```diff
@@ -100,45 +98,49 @@ void MStripHit::Clear()
...
     double energy, energy_res;
     char pos_strip;
     unsigned int flags;
-    int N = sscanf(&line[3], "%u %c %u %d %lf %lf %lf %lf %lf %u",
+    int N = sscanf(&line[3], "%u %c %u %d %lf %lf %lf %lf %u",
                    &det_id, &pos_strip, &strip_id, &has_triggered,
-                   &timing, &un_adc, &adc, &energy, &energy_res, &flags);
-    if (N != 10) {
-      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: malformed SH line"<<endl;
+                   &timing, &adc, &energy, &energy_res, &flags);
```
ckierans, line 122: It looks like we're missing tac and timing resolution here?

src/MStripHit.cxx
```diff
@@ -100,45 +98,49 @@ void MStripHit::Clear()
...
+    if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: line too short with length "<<Line.Length()<<endl;
     return false;
   }
 
   if (line[0] == 'S' && line[1] == 'H') {
     unsigned int det_id, strip_id;
     int has_triggered;
-    double timing, un_adc, adc;
+    double timing, adc;
     double energy, energy_res;
```
ckierans, line 117: Seems weird to combine timing and adc here. Maybe instead:

double adc, tac;
double energy, energy_res;
double timing, timing_res;

Then after Issues #176 is resolved, the top line can change to int.

See my comment below. Is there a reason we don't have tac and timing_res here, or is that an oversight?

src/MStripHit.cxx
```diff
@@ -100,45 +98,49 @@ void MStripHit::Clear()
...
-    if (N != 10) {
-      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: malformed SH line"<<endl;
+                   &timing, &adc, &energy, &energy_res, &flags);
+    if (N != 9) {
+      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: malformed SH line with "<<N<<" fields instead of 9"<<endl;
       return false;
     }
     if (pos_strip != 'l' && pos_strip != 'h') {
-      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: unknown detector face: "<<pos_strip<<endl;
+      if (g_Verbosity >= c_Error) cout<<"Error in MStripHit::Parse: unknown detector face '"<<pos_strip<<"' (expected 'l' or 'h')"<<endl;
```
ckierans: I would say "side" instead of "face"

zoglauer: Done

src/MStripHit.cxx
zoglauer: We have to redo all dat reading and writing - there are several issues. See #179
We have to do the fix them all at once. 
Thus I just fixed emergency things there at the moment.

src/MStripHit.cxx
zoglauer: We have to redo all dat reading and writing - there are several issues. See #179
We have to do the fix them all at once in all DAT classes.
Thus I just fixed emergency things there at the moment.

include/MStripHit.h
zoglauer: Done

src/MStripHit.cxx
zoglauer: We have to redo all dat reading and writing - there are several issues. See #179
We have to do the fix them all at once in all DAT classes.
Thus I just fixed emergency things there at the moment.

Is anybody using this? - then I can do an emergency fix.

src/MReadOutAssembly.cxx
zoglauer: The temperature has to come in differently, if we want to do a temperature dependent calibration in the future with melinator, if we don't store the info in the hit itself.
But there is a chance that we will have to to a temperature dependent energy calibration.

src/MStripHit.cxx
zoglauer: Done

include/MStripHit.h
zoglauer: See issue #176

include/MStripHit.h
zoglauer: Removed

include/MStripHit.h
zoglauer: Removed

include/MStripHit.h
zoglauer: #176

include/MStripHit.h
zoglauer: Removed - we do have to think about how we will handle temperature dependent energy calibration in the future... in case it is temperature dependent...

include/MStripHit.h
zoglauer: Done

include/MStripHit.h
zoglauer: Removed

include/MStripHit.h
zoglauer: is 0 ever a good timing? It is, right? Since we can have positive and negative values. Thus we need something indicating that the thing has been set.

fhagemann: I would say 0 is never a good timing, because „timing“ is the calibrated time in nanoseconds between starting the TAC clock and ending it. From ENABLE rise to fall should always be minimum 2000ns for strips with fast timing. Not sure if we expect negative slow timings..

include/MStripHit.h
NicoleRodriguezCavero: Answer to: Do we need the `CalibratedTiming` flag?
For the TACcut module `HasCalibratedTiming` is redundant and a left-over from previous versions of the module; we can achieve the same cuts using `HasFastTiming` + `IsGuardRing`. I can double check if `HasCalibratedTiming` is being used anywhere else and, if not, update the TACcut module.

include/MStripHit.h
```diff
@@ -83,7 +83,7 @@ class MStripHit
   //! TODO: Switch to unsigned int: Issue #176
   //! Set the measured ADC units of the strip
   void SetADCUnits(double ADCUnits) { m_ADCUnits = ADCUnits; }
-  //! Return measured the ADC units of the strip
+  //! Return the measured the ADC units of the strip
```
fhagemann:
```suggestion
  //! Return the measured ADC units of the strip
```

fhagemann (commented):
Tiny typo

PR #173, opened 2026-06-29
Remove nearest neighbor hack

### LINKED ISSUE #93: Remove nearest neighbor hack
Remove nearest neighbor hack from trigger modules once nearest neighbor is implemented. 

Adding this as an issue so we can keep track of it.

parshadkp:
This will remove the nearest neighbor hack that was introduced a while ago to calculate the correct dead time in the strip trigger module. The change is simple but I will make this "ready to review" once PR #165 is merged. This will also close Issue #93.


fhagemann:
This might probably also need the implementation of strip thresholds + some nearest neighbor logic on top of #165, but I can take care of that once #165 is approved.

include/MSubModuleStripTrigger.h
```diff
@@ -122,9 +122,6 @@ class MSubModuleStripTrigger : public MSubModule
   //! Process strip hits for deadtime and trigger determination
   bool ProcessStripHits(MReadOutAssembly* Event);
 
-  //! Helper function for getting count rate (including nearest neighbor)
-  bool CountRate(vector<int> ASICChannels, vector<double> CountTime);
```
ckierans: What's the motivation for removing this function?

src/MSubModuleStripTrigger.cxx
```diff
@@ -159,27 +158,8 @@ double MSubModuleStripTrigger::CalculateASICDeadtime(vector<int> ASICChannels)
...
-      // Edge case: If ID is 31 or 63, add the previous channel and the channel itself
-      ASICChannelsSet.insert(ID - 1);
-      ASICChannelsSet.insert(ID);
-    } else {
-      // General case: Add the previous channel, the channel itself, and the next channel
-      ASICChannelsSet.insert(ID - 1);
-      ASICChannelsSet.insert(ID);
-      ASICChannelsSet.insert(ID + 1);
-    }
+    ASICChannelsSet.insert(ID);
```
ckierans: That's much easier :)

ckierans (commented):
Looks good, Parshad. I'm fine with you removing the CountRate function, since it seems like something that doesn't quite belong here as a diagnostic, but I just wanted to check your motivation for removing it.

include/MSubModuleStripTrigger.h
parshadkp: Hi @ckierans, the count rate is still printed out with c_Info. This was an outdated function that printed out the count rate using the NN hack. Since the NN are included from charge transport, this is no longer needed.

parshadkp:
Hi @ckierans, I think this looks good to merge now but let me know if you have any other comments.

fhagemann:
But this also has the code from PR190 in it now?

parshadkp:
> But this also has the code from PR190 in it now?

Sorry, I must have accidentally pushed my local changes when I was testing. Should be fixed now.

fhagemann (approved):
Then let's merge this PR together with #190 once that is ready :)

PR #144, opened 2026-05-13
[DEE] Reduce code duplication in parsing energy calibration file

fhagemann:
This is my attempt at reducing the code duplication in `MSubModuleStripReadout`.
Currently, we have two implementations of reading an ecal file; in `MModuleEnergyCalibration` and `MSubModuleStripReadout`. The only difference between them seems to be that the energy resolution in `MModuleEnergyCalibration` is parsed in units of `σ` by dividing by 2.355 already in the `TF1` function, whereas in `MSubModuleStripReadout` it is parsed in units of FWHM in the `TF1` function, and then converted to `σ` later in the code.

By exporting the code to read the energy calibration file to a dedicated function, we can avoid the code duplication.

I ran the code twice, with the current `develop/em` branch and the updated code in this PR to cross-check that I obtain the same results and that the energy resolution is handled in the correct units (`σ`/FWHM) :
image

PR #181, opened 2026-07-15
Trapping correction branch

sophieehaight:
Contains a new module for applying depth-based charge trapping correction to individual hit energies along with modules with GUI options and GUI expos. The branch also includes an app for characterizing trapping with Cs-137 data. The trapping correction module requires a csv parameter file as input: [detector_0_trapping_parameters.csv](https://github.com/user-attachments/files/30068579/detector_0_trapping_parameters.csv)


src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,684 @@
...
+  } 
+
+  else {
+    // If more complicated than the above cases, return 4 for now.
+    // TODO: Handle more complicated charge distributions.
+    return_value = 4;
+  }
+
+  return return_value;
+}
```
fhagemann, line 580: Is this identical to `MModuleDepthCalibration::GetHitGrade`?
If so, we could avoid code duplication here by calling the original `MModuleDepthCalibration::GetHitGrade` instead.

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,684 @@
...
+          m_DetectorIDs.push_back(DetID);
+          m_Detectors[DetID] = det;
+        } else {
+          if (g_Verbosity >= c_Error) {
+            cout<<"ERROR in MModuleTrappingCorrection::Initialize: Found a duplicate detector: "<<det_name<<endl;
+          }
+        }
+      }
+    }
+  }
```
fhagemann, line 140: This is also done in `MModuleDepthCalibration`, so maybe we can export this to a function of `MModuleDepthCalibration` and call that new function here as well to avoid code duplication.
Might also be relevant for #182

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,684 @@
...
+
+      // Handle different grades differently  
+      // Get the position from the depth cal. If error is thrown, record and no depth.
+
+      // GRADE=-1 is an error. Break from the loop and continue.
+      if (Grade < 0){
+        H->SetNoDepth();
+      } else if (Grade > 4) { // GRADE=5 is some complicated geometry with multiple hits on a single strip. GRADE=6 means not all strips are adjacent.
+        H->SetNoDepth();
+      } else { // If the Grade is 0-4, we can handle it.
```
fhagemann, line 209: Isn't this already done in `MModuleDepthCalibration` (setting `H->SetNoDepth()`)?
https://github.com/cositools/nuclearizer/blob/1a1b3eedef8b0afb964fb72d9bed19190fe352fc/src/MModuleDepthCalibration.cxx#L199-L216

If this charge trapping module is called AFTER the depth calibration, not sure if this is needed here (?)

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,684 @@
...
+      MaxEnergy = Energy;
+    }
+  }
+  if (TotalEnergy == 0) {
+    EnergyFraction = 0;
+  } else {
+    EnergyFraction = MaxEnergy/TotalEnergy;
+  }
+  return MaxStrip;
+}
```
fhagemann, line 338: This also exists in `MModuleDepthCalibration`, so no need to duplicate ;)

include/MModuleTrappingCorrection.h
```diff
@@ -0,0 +1,158 @@
...
+  // protected methods:
+ protected:
+  //! Returns the strip with most energy from vector Strips, also gives back the energy fraction
+  MStripHit* GetDominantStrip(std::vector<MStripHit*>& Strips, double& EnergyFraction);
+  
+  // //! Retrieve the appropriate Depth values given the DetID
+  // vector<double> GetDepth(int DetID);
+
+  //! Determine the Grade (geometry of charge sharing) of the Hit
+  int GetHitGrade(MHit* H);
```
fhagemann, line 96: This can also be removed (maybe check #144 for how we avoided code duplication in the DEE by reusing function from the energy calibration module).

fhagemann (commented):
I had a quick glimpse, and it seems like you're duplicating some function from the `MModuleDepthCalibration` to use here. Consider avoiding that code duplication by calling the original functions directly.

PR #183, opened 2026-07-24
Implementing NN into Strip Pairing

julianmgerber:
Previously, there was nothing about NN in the strip pairing, so if someone tried the strip pairing with NN included then it'd try to group NN + triggered strips together and add in all the energy on the NN strips. Perhaps eventually we'll want to add in some of the NN energy, but for now I thought it made sense to keep them separate.

Changes:
- Separate out NN/triggered strips so that only the triggered strips are actually paired and grouped into hits
- After the pairing, NN strips are linked to their neighboring hit through a new function: AddNearestNeighborStripHit
- If it's ambiguous which hit a NN strip hit should be associated with (like if there are two hits separated by only one strip), then the NN strip is linked to both

Questions:
- Should there be a way to read out the NN strip hits into a .dat file (since right now, only the actual triggered SH making up a hit will be read out)?
       - Update 7/28: Reading out NN strip hits to .dat file as "NN" rather than "SH"
- Should there be a separate flag for those ambiguous NN strip hits that are linked to two different hits?
       - Update 7/28: New IsAmbiguousNearestNeighbor flag added to MStripHit class to flag NNs that are linked to multiple hits


src/MStripHit.cxx
```diff
@@ -161,19 +161,36 @@ void MStripHit::AddOrigins(vector<int> Origins)
...
-   <<m_ADCUnits<<" "
-   <<m_Energy<<" "
-   <<m_EnergyResolution<<" "
-   <<MakeFlags()<<endl;
+  if (m_ReadOutElement->IsNearestNeighbor() == true) {
+    S<<"NN "
+    <<m_ReadOutElement->GetDetectorID()<<" "
+    <<((m_ReadOutElement->IsLowVoltageStrip() == true) ? "l" : "h")<<" "
+    <<m_ReadOutElement->GetStripID()<<" "
+    <<m_HasTriggered<<" "
```
fhagemann, line 175: Just wondering: is `m_HasTriggered` correlated with the strip being a `NN`?
Maybe this here might be redundant then, as we can "reconstruct" NN from SH that did not trigger.

ckierans: Andreas has removed the m_HasTriggered flag in PR #171 .

fhagemann: `m_HasTriggered` seems to not have been removed in #171.

julianmgerber: From what I can tell, m_HasTriggered isn't actually assigned anywhere. Like in the .dat file output, all the strip hits (triggered or otherwise) just have 0 for the HasTriggered parameter.

fhagemann: Is this when you read data from HDF5? 
Maybe, the HDF5 reader module does not assign values to `m_HasTriggered`? (but should?)

fhagemann: Yes, looks like the HDF5 reader never assigns a `m_HasTriggered` value:
https://github.com/cositools/nuclearizer/blob/18e1de5a335be5e4f792cffb4686ae4403a9c43f/src/MModuleLoaderMeasurementsHDF.cxx#L711-L726

include/MModuleStripPairingMultiRoundChiSquare.h
```diff
@@ -87,7 +87,13 @@ class MModuleStripPairingMultiRoundChiSquare : public MModule
     float ChargeTrappingCorrection(unsigned int d, const vector<vector<MStripHit*>>& StripHits);
     
   //! Divide an event's strip hits by detector and LV/HV side
-    vector<vector<vector<MStripHit*>>> CollectStripHits(MReadOutAssembly* Event);
+    tuple<vector<vector<vector<MStripHit*>>>, bool> CollectStripHits(MReadOutAssembly* Event);
```
ckierans: Are these now only triggered strip hits? Maybe make that clear in the comment?

include/MHit.h
```diff
@@ -52,9 +52,13 @@ class MHit
   //! Return strip hit i or nullptr if i is out of bounds
   //! Ownership stays elsewhere
   MStripHit* GetStripHit(unsigned int i);
+  //! Return nearest neighbor strip hit i or nullptr if i is out of bounds
+  MStripHit* GetNearestNeighborStripHit(unsigned int i);
```
ckierans: I'm wondering about any potential confusion/overlap with NN strip hits listed here vs in the StripHit list that then have the IsNearestNeightbor flag. Do we expect NN strip hits to also be listed in the standard GetStripHit list or is that now exclusively for triggered strips? If so, let's make that clear.

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -234,40 +234,93 @@ float MModuleStripPairingMultiRoundChiSquare::ChargeTrappingCorrection(unsigned
...
+        vector<vector<MStripHit*>> List; // list of sides, list of hits
+        List.push_back(vector<MStripHit*>()); // LV
+        List.push_back(vector<MStripHit*>()); // HV
+        List[Side].push_back(SH);
+        StripHits.push_back(List);
+        DetectorIDs.push_back(SH->GetDetectorID());
       }
     }
+    else {
+      IncludingNearestNeighbors = true;
```
ckierans: I'm not sure I understand how this boolean is supposed to be used. At the beginning, you define Including NearestNeighbors as false. But if any of the SH are found to have the IsNearestNeighbor flag set as true, then you change the variable here? 

We have the boolean SetIncludeNearestNeighbors in the MModuleLeaderMeasurementHDF module, which theoretically should be the same as this boolean here, except you're deriving it from the data, and not the GUI.

include/MModuleStripPairingMultiRoundChiSquare.h
ckierans: Yes, I see below these are only the triggered strip hits. You should update the comment to reflect this change.

"//! Divide an event's triggered strip hits by detector and LV/HV side"

include/MHit.h
ckierans: Is this now only triggered strip hits? Or both triggered and untriggered?

src/MModuleStripPairingMultiRoundChiSquare.cxx
```diff
@@ -711,6 +764,74 @@ bool MModuleStripPairingMultiRoundChiSquare::CreateHits(unsigned int d, MReadOut
...
+  vector<MStripHit*> AssignedNeighbors; // List of all the assigned NN strip hits, in order to check if NNs are double counted
+  
+  for (unsigned int h = 0; h < Event->GetNHits(); h++) {
+    vector<vector<int>> StripIDs; // list of sides, list of strips
+    StripIDs.push_back(vector<int>()); // LV
+    StripIDs.push_back(vector<int>()); // HV
+    bool AssignedDetector = false;
+    int DetectorID; // Define detector ID where hit took place
+    for (unsigned int sh = 0; sh < Event->GetHit(h)->GetNStripHits(); sh++) {
+      // Collect all the strip hits in a hit and split them by side
```
ckierans: I assume that whenever this function is called, the Hit only has triggered strips assigned to it. So this GetNStripHits will only return the number of triggered strip hits. Make that clear in the comment?

Once you assign Nearest Neighbor strip hits to the event, do you expect GetNStripHits to count all associated strip hits, or only triggered strips?

src/MStripHit.cxx
ckierans: Okay, I spoke with confidence when I didn't actually check the PR #171 code! Sorry about that. 

I'm pretty sure we agreed we didn't need the HasTriggered boolean because it is completely redundant with (but opposite to) the IsNearestNeighbor flag. As far as I can tell, no one is using this flag anywhere so I'd vote for removing it entirely to avoid confusion going forward.

ckierans: But that doesn't need to happen in this PR, to be clear.

ckierans (commented):
Looks good, Julian. I think we need to be a bit more careful about being clear when the StripHit list is only triggered strips or also includes NN strip hits as well. Perhaps this can just be better comments, but we might consider changing some of the names to include "TriggeredStripHits" so there's no confusion.

src/MStripHit.cxx
fhagemann: I agree (both removing `HasTriggered` and that it doesn't have to happen in this PR).

fhagemann: `m_HasTriggered` was replaced with (the inverse of) `m_IsNearestNeighbor` in #202 (which was also merged), so maybe instead of renaming this to `NN`, we can use the `m_IsNearestNeighbor` flag instead (?)

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: GetNStripHits should only count triggered SH since the triggered SH get added to m_StripHits while the NNs get added to m_NearestNeighborStripHits

include/MHit.h
julianmgerber: GetStripHit from the MHit class should only be triggered strip hits. However, the GetStripHit function in MReadOutAssembly (for the whole event) would have both triggered and NN strips. I added some comments to clarify

src/MModuleStripPairingMultiRoundChiSquare.cxx
julianmgerber: Felix pointed out that the HDF loader module isn't a required prerequisite for the strip pairing, so that boolean might not always be set (like if someone is just loading an ROA file, for instance). I moved this flag though to a different place and added some comments, so hopefully it's clearer.

include/MHit.h
julianmgerber: Should just be triggered. I added a comment to clarify.

julianmgerber:
@ckierans I updated the comments and variable names within the strip pairing to try and make it clearer where there's triggered versus NN strip hits.

For the streamDAT function, I left it reading out as "NN" instead of "SH" because I think it's a little helpful to immediately see all the nearest neighbors. But this is totally redundant with the "IsNearestNeighbor" flag, so I can change that if people prefer differently.

ckierans:
This looks good, @julianmgerber! Thanks for adding all of those clarifying comments and more detailed variable names. It's much easier to follow. I like the "NN" in the .dat file.

PR #184, opened 2026-07-24
Trapping correction module and app

### LINKED ISSUE #41: Implement Charge Trapping correction
Based on Sean’s Charge Trapping App with Am-241, the Charge Trapping correction can be linear based on the boundary conditions found in the app:

Image

Slack message from Sean September 2025: “My intention was to just do a linear correction so we don't have to do any simulations or anything. Since we're anticipating a small effect for the undamaged detectors, I figured the linear approximation is probably sufficient.”

With this, we need to also determine the mechanics of running the data through the first steps of Nuclearizer, implementing the charge trapping correction, refitting in Melinator, and then running the data through Nuclearizer again.

sophieehaight:
Contains a new module for applying depth-based charge trapping correction to individual hit energies along with modules with GUI options and GUI expos. The branch also includes an app for characterizing trapping with Cs-137 data. The trapping correction module requires a csv parameter file as input:
[detector_0_trapping_parameters.csv](https://github.com/user-attachments/files/30357422/detector_0_trapping_parameters.csv)
This file can also be found in resource/dee


apps/TrappingCorrectionCs137.cxx
```diff
@@ -781,37 +753,8 @@ TF1* TrappingCorrectionCs137::GeneratePhotopeakFunction()
...
-
-//   return CTDFunction;
-// }
-
-
-////////////////////////////////////////////////////////////////////////////////
-
-
-TrappingCorrectionCs137* g_Prg = 0;
-int g_NInterruptCatches = 1;
```
fhagemann: You're getting an error now because this is used latter in the app and is now undefined.
Can you put those back in or adjust the code latter in the app to not rely on these anymore?

src/MModuleTrappingCorrection.cxx
fhagemann: Something like this is also done in `MModuleDepthCalibration`, so maybe you can also remove this code here and retrieve the information on the thicknesses from `m_DepthCalibration` as well?

sophieehaight: I fixed this and added a public function for getting detector IDs in the depth cal header

include/MModuleTrappingCorrection.h
```diff
@@ -0,0 +1,161 @@
...
+
+
+  // protected methods:
+ protected:
+
+  //! Load in the specified SimCCE file
+  bool LoadSimCCEFile(MString FName);
+
+  //! Get the Sim-based corrected energy given the CTD value, uncorrected energy, and the sorted Sim CCE values
+  double GetSimBasedCorrectedEnergy(double ctd_val, double uncorrected_energy, const std::vector<double>& sim_cce_sorted, double paramA, double paramB, double paramC);
```
fhagemann: This seems to have changed:

```suggestion
  //! Get the Sim-based corrected energy given the depth value, uncorrected energy, and the sorted Sim CCE values
  double GetSimBasedCorrectedEnergy(double depth_val, double uncorrected_energy, const std::vector<double>& sim_cce_sorted, double paramA, double paramB, double paramC);
```

resource/dee/dummy_trapping_parameters_singledetector.csv
fhagemann, line 1: In this dummy file, the CCE curves for HV and LV side are exactly equal.
Is that expected?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,570 @@
...
+
+        // Get the dominant strip for the hit and its energy fraction for both LV and HV sides
+        //s.t we cna determine the depth anf energy to correct
+        double LVEnergyFraction;
+        double HVEnergyFraction;
+        MStripHit* LVSH = m_DepthCalibration->GetDominantStrip(LVStrips, LVEnergyFraction); 
+        MStripHit* HVSH = m_DepthCalibration->GetDominantStrip(HVStrips, HVEnergyFraction); 
+
+        // Get the position value (assumed from event/hit context H)
+        double depth_val = H->GetPosition().GetZ();
```
fhagemann: `depth_val` is in units of `cm`, and then passed to `GetSimBasedCorrectedEnergy`...

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,570 @@
...
+      }
+    } 
+    else if (ValidLineCount == 1) {
+      // Skip the second line which is the column header (z_depth_mm,CCE_HV)
+      ValidLineCount++;
+    } 
+    else {
+      // Read the rest of the rows into your depth and CCE HV arrays
+      if (Tokens.size() == 3) {
+        m_Depths.push_back(Tokens[0].ToDouble());
```
fhagemann, line 382: `m_Depths` is read from the file in units of `mm`

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,570 @@
...
+  return true;
+}
+
+
+/////////////////////////////////////////////////////////////////////////////////
+
+double MModuleTrappingCorrection::GetSimBasedCorrectedEnergy(double depth_val, double uncorrected_energy, const std::vector<double>& sim_cce_sorted, double paramA, double paramB, double paramC) {
+
+  // Look up the simulation CCE baseline using the interpolate function
+  double cce_base = Interpolate(depth_val, m_Depths, sim_cce_sorted);
```
fhagemann: And then `m_Depths` (in units of mm) is evaluated at the `depth_val` (in units of cm), which only evaluates the CCE curve in the region -0.75mm to 0.75mm, instead of -0.75cm to 0.75cm.

fhagemann (changes requested):
It looks like the units of `depth_val` (mm) and `m_Depths` (cm) do not match..

resource/dee/dummy_trapping_parameters_singledetector.csv
fhagemann: After the last commit, the CCE curves look more reasonable now:

Image

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,570 @@
...
+
+  // Evaluate the physical trapping function model using class global popt variables
+  double expected_centroid_scaled = paramA * (1.0 - paramB * (1.0 - cce_base)) * (1.0 - paramC * (1.0 - cce_base));
```
fhagemann: I had a look at the paper ([Boggs et al. (2023)](https://doi.org/10.1016/j.nima.2023.168756)) that this charge trapping correction is based on:

Image

Image

I'm a bit confused by what is stated in the paper and what's done here:
- First, it looks like there should be four parameters (A_e, A_h, B, C), whereas you list six (A_LV, A_HV, B_LV, B_HV, C_LV, C_HV), and the parameter pairs (B_LV, B_HV) and (C_LV, C_HV) are each very different.
- It looks like there should be 4 CCE curves: 
  - CCE_HV with only electron trapping, 
  - CCE_HV with only hole trapping, 
  - CCE_LV with only electron trapping, 
  - CCE_LV with only hole trapping. 
  
  Your example file lists only two CCE curves, I assume they are:
  - CCE_HV with both electron and hole trapping,
  - CCE_LV with both electron and hole trapping.

  Then you're using the same CCE curves twice in the equation to determine the `expected_centroid_scaled` here, which I don't think is what the original method listed in the paper intended.

include/MModuleTrappingCorrection.h
sophieehaight: I updated this parameter to reflect that the CCE / energy correction is being evaluated in depth space rather than CTD

src/MModuleTrappingCorrection.cxx
sophieehaight: Hey Felix, in accordance with our slack discussions about this I just pushed an updated parameter file with the raw simulated CCE curves (HV and LV curves showing the effects of electron and hole trapping separately -- HV_CCE_e, HV_CCE_h, LV_CCE_e, LV_CCE_h) and the four fitted trapping parameters (A_hv, A_lv, B, C).

src/MModuleTrappingCorrection.cxx
sophieehaight: This should be updated on the new trapping parameter file s.t CCE depths (m_Depths) are read in in units of mm to match the depths we are getting from data (depth_val)

fhagemann: Not sure if I'm misunderstanding, but I feel like the problem still persists:
`depth_val` is in units of `cm`, whereas the depths in your CCE files are in `mm`.

So if, you have an event at a depth of `6mm`, `depth_val` would be `0.6` (`6mm = 0.6cm`), and then you would evaluate the CCE curves at `0.6` (which would correspond to `0.6mm`).

Maybe define the depths in your CCE files also in `cm`, just like we do it for the depth splines?
https://github.com/cositools/nuclearizer/blob/develop/em/resource/dee/dummy_15.2mm_spline.csv

sophieehaight: I'm a bit confused about the units produced by GetPosition() because when I readout the values for depth_val the majority are > 1.5  which makes it seem like they should be in units of mm because otherwise they would be outside of the dimensions of the detector. However, I do see that the depth spline files are read in in units of cm so I would assume that the depth cal happens all in cm.

fhagemann: We would need to check if `GetPosition` is the position in absolute coordinates, or in relative coordinates within the detector. Also: if it's in relative coordinates, the values should span from -0.75 to 0.75 if in cm (or -7.5 to 7.5 if in mm).

Are you generating this for the STTC? If so, in the absolute coordinates, the depth is actually aligned with the `Y`-axis, so that could explain why you would get `Z` positions exceeding the expected detector dimensions.

fhagemann: Yeah, looks like the depth calibration adds the `GlobalPosition` instead of the `LocalPosition`
(whereas your charge trapping requires knowing the `LocalPosition`). Maybe it makes sense to store both in the depth calibration?

https://github.com/cositools/nuclearizer/blob/18e1de5a335be5e4f792cffb4686ae4403a9c43f/src/MModuleDepthCalibration.cxx#L361-L369

fhagemann: Or maybe we can infer this with `GetPositionInSensitiveVolume` (?)

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,571 @@
...
+      }
+    } 
+    else if (ValidLineCount == 1) {
+      // Skip the second line which is the column header (z_depth_mm,CCE_HV)
+      ValidLineCount++;
+    } 
+    else {
+      // Read the rest of the rows into your depth and CCE HV arrays
+      if (Tokens.size() == 3) {
+        m_Depths.push_back(Tokens[0].ToDouble());
```
fhagemann, line 382: I did this in #187, to have all length units in the charge trapping correction consistently in `cm`.
```suggestion
        m_Depths.push_back(Tokens[0].ToDouble() / 10.0);
```

src/MModuleTrappingCorrection.cxx
sophieehaight: To address this I created a member in the MHit class for the local position and I updated the depth calibration to record the global and local position of each hit with a calibrated depth. Now in the trapping correction, I access the Z coordinate of this local position instead (which looks consistent with units of cm rather than mm).

src/MModuleTrappingCorrection.cxx
```diff
@@ -201,13 +201,14 @@ bool MModuleTrappingCorrection::AnalyzeEvent(MReadOutAssembly* Event)
         MStripHit* LVSH = m_DepthCalibration->GetDominantStrip(LVStrips, LVEnergyFraction); 
         MStripHit* HVSH = m_DepthCalibration->GetDominantStrip(HVStrips, HVEnergyFraction); 
 
-        // Get the position value (assumed from event/hit context H)
-        double depth_val = H->GetPosition().GetZ();
+        // Get the position value in mm(assumed from event/hit context H)
```
fhagemann, line 204: Is `in mm` correct here? I would assume all output of `GetX`, `GetY` and `GetZ` to always be in units of `cm`

apps/TrappingCorrectionAm241.cxx
fhagemann: You added some `cout` lines to this: should this stay, or was this primarily for debugging and can be removed?

apps/TrappingCorrectionCs137.cxx
```diff
@@ -0,0 +1,833 @@
+/* 
+ * TrappingCorrectionCs137.cxx
+ *
+ *
+ * Copyright (C) by Sean Pike.
```
fhagemann: Want to take credit for this as well? ^^

apps/TrappingCorrectionCs137.cxx
```diff
@@ -0,0 +1,833 @@
...
+//! Parse the command line
+bool TrappingCorrectionCs137::ParseCommandLine(int argc, char** argv)
+{
+  ostringstream Usage;
+  Usage<<endl;
+  Usage<<"  Usage: TrappingCorrection <options>"<<endl;
+  Usage<<"    General options:"<<endl;
+  Usage<<"         -i:   input file name (.hdf5 or .txt with list of hdf5s)"<<endl;
+  Usage<<"         --emin:   minimum Event energy (default 30 keV)"<<endl;
+  Usage<<"         --emax:   maximum Event energy (default 5000 kev)"<<endl;
```
fhagemann: Later it looks like the defaults for `--emin` and `--emax` are actually 600 and 700keV, respectively. Can you cross-check all default values in this help message?

apps/TrappingCorrectionCs137.cxx
```diff
@@ -0,0 +1,833 @@
...
+  if ((InputFile.GetSubString(InputFile.Length() - 4)) == "hdf5") {
+    HDFNames.push_back(InputFile);
+    cout << "hdf names loaded correctly" << endl;
+  } else if ((InputFile.GetSubString(InputFile.Length() - 3)) == "txt") {
+    cout << "Reading input file " << InputFile << endl;
+    cout << "WARNING: When passing a list of files, ensure that you have chosen the correct HDF5 continuous reading mode. Use the --nocontinue option to suppress continuous file reading." << endl;
+    MFile F;
+    if (F.Open(InputFile) == false) {
+      cout << "Error: Failed to open input file." << endl;
+    } else {
```
fhagemann: Are there examples for a `txt` input file?

apps/TrappingCorrectionCs137.cxx
fhagemann: Do you have an example command to show how to run this app?

include/MHit.h
```diff
@@ -178,6 +186,10 @@ class MHit
   //! Position resolution of the hit
   MVector m_PositionResolution;
 
+
+  //! Local/raw position resolution of the hit
+  MVector m_LocalPosition;
```
fhagemann: This has nothing to do with resolution..
This is the position of the hit in local detector coordinates:

```suggestion
  //! Position of the hit in local detector coordinates
  MVector m_LocalPosition;
```

include/MModuleTrappingCorrection.h
```diff
@@ -0,0 +1,163 @@
...
+ protected:
+
+  //! Load in the specified SimCCE file
+  bool LoadSimCCEFile(MString FName);
+
+  //! Get the Sim-based corrected energy given the CTD value, uncorrected energy, and the sorted Sim CCE values
+  double GetSimBasedCorrectedEnergy(double ctd_val, double uncorrected_energy, const std::vector<double>& sim_cce_sorted_e, const std::vector<double>& sim_cce_sorted_h, double paramA, double paramB, double paramC);
+
+  //! Interpolate a value given x, xp, and fp
+  double Interpolate(double x, const std::vector<double>& xp, const std::vector<double>& fp);
```
fhagemann: I wonder if such a function already exists to interpolate the CTD-depth spline, and if this can be reused here to avoid code duplication.

resource/dee/dummy_trapping_parameters_singledetector.csv
```diff
@@ -0,0 +1,153 @@
+1.0009264205609762,1.00084829470918,1.288689028046797,0.3010529785474152
+z_depth_cm,e_CCE_HV,h_CCE_HV,e_CCE_LV,h_CCE_LV
```
fhagemann: Maybe add this as comments, so that it's clear what all of these values are?

```suggestion
# A_HV,A_LV,B,C
1.0009264205609762,1.00084829470918,1.288689028046797,0.3010529785474152
# z_depth_cm,e_CCE_HV,h_CCE_HV,e_CCE_LV,h_CCE_LV
```

Will this format allow to concatenate simulated CCEs and parameters for multiple detectors?

src/MGUIOptionsLoaderMeasurementsHDF.cxx
```diff
@@ -85,12 +85,6 @@ void MGUIOptionsLoaderMeasurementsHDF::Create()
     dynamic_cast<MModuleLoaderMeasurementsHDF*>(m_Module)->GetFileNameStripMap());
   m_FileSelectorStripMap->SetFileType("Strip map file", "*.map");
   m_OptionsFrame->AddFrame(m_FileSelectorStripMap, LabelLayout);
-  
-  // Nearest neighbor checkbox
-  m_IncludeNearestNeighbor = new TGCheckButton(m_OptionsFrame, "Include Nearest Neighbors");
-  m_IncludeNearestNeighbor->SetOn(dynamic_cast<MModuleLoaderMeasurementsHDF*>(m_Module)->GetIncludeNearestNeighbor());
-  m_IncludeNearestNeighbor->Associate(this);
-  m_OptionsFrame->AddFrame(m_IncludeNearestNeighbor, LabelLayout);
```
fhagemann: Why was this removed?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,628 @@
...
+{
+  m_SimCCEFileIsLoaded = LoadSimCCEFile(m_SimCCEFile);
+  if (m_SimCCEFileIsLoaded == false) {
+    return false;
+  }
+  
+  MSupervisor* S = MSupervisor::GetSupervisor();
+  m_EnergyCalibration = (MModuleEnergyCalibration*) S->GetAvailableModuleByXmlTag("EnergyCalibration");
+  if (m_EnergyCalibration == nullptr) {
+    cout << "MModuleTrappingCorrection: couldn't resolve pointer to Energy Calibration Module... need access to this module for energy resolution lookup!" << endl;
```
fhagemann: Can you wrap the `cout` in `g_Verbosity` checks (also elsewhere)?

```suggestion
    if (g_Verbosity >= c_Error) {
      cout << "MModuleTrappingCorrection: couldn't resolve pointer to Energy Calibration Module... need access to this module for energy resolution lookup!" << endl;
    }
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,628 @@
...
+}
+
+
+/////////////////////////////////////////////////////////////////////////////////
+
+
+bool MModuleTrappingCorrection::AnalyzeEvent(MReadOutAssembly* Event) 
+{
+  if (Event->GetGuardRingVeto() == true) {
+    // Right now we cannot use events w GR veto
```
fhagemann: Is this because those events have no depth-information?
Will this change in the future?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,628 @@
...
+      }
+    } 
+    else if (ValidLineCount == 1) {
+      // 2. Print column headers line
+      cout << "Loaded Column Headers: ";
+      for (size_t i = 0; i < Tokens.size(); ++i) {
+        cout << Tokens[i] << (i + 1 < Tokens.size() ? " | " : "\n");
+      }
+      ValidLineCount++;
+    }
```
fhagemann: Enclose this is `g_Verbosity` check (to me this is info):

```suggestion
    else if (ValidLineCount == 1) {
    
      // 2. Print column headers line
      if (g_Verbosity >= c_Info) {
        cout << "Loaded Column Headers: ";
        for (size_t i = 0; i < Tokens.size(); ++i) {
          cout << Tokens[i] << (i + 1 < Tokens.size() ? " | " : "\n");
        }
      }
      ValidLineCount++;
    } 
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,628 @@
...
+        // Print the first 3 data rows to terminal for verification
+        if (ValidLineCount <= 4) {
+          cout << "Data Row " << (ValidLineCount - 1) << ": ";
+          for (size_t i = 0; i < Tokens.size(); ++i) {
+            cout << Tokens[i] << (i + 1 < Tokens.size() ? " | " : "\n");
+          }
+          if (ValidLineCount == 4) {
+            cout << "-----------------------------------\n" << endl;
+          }
+        }
```
fhagemann: This to are is also either info, or even debug print statements:

```suggestion
      // 3. Read the 5 data columns into depth and CCE arrays
      if (Tokens.size() == 5) {
        // Print the first 3 data rows to terminal for verification
        if (g_Verbosity >= c_Info) {
          if (ValidLineCount <= 4) {
            cout << "Data Row " << (ValidLineCount - 1) << ": ";
            for (size_t i = 0; i < Tokens.size(); ++i) {
              cout << Tokens[i] << (i + 1 < Tokens.size() ? " | " : "\n");
            }
            if (ValidLineCount == 4) {
              cout << "-----------------------------------\n" << endl;
            }
          }
        }
```

or remove completely

```suggestion
```

src/MModuleTrappingCorrection.cxx
fhagemann: `m_Depths` is read from the file in units of `cm` now, so this seems to be resolved.

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,628 @@
...
+
+        m_Depths.push_back(Tokens[0].ToDouble());
+        m_CCEs_HV_e.push_back(Tokens[1].ToDouble());
+        m_CCEs_HV_h.push_back(Tokens[2].ToDouble());
+        m_CCEs_LV_e.push_back(Tokens[3].ToDouble());
+        m_CCEs_LV_h.push_back(Tokens[4].ToDouble());
+        
+        ValidLineCount++;
+      } else {
+        cout << "ERROR in LoadSimCCEFile: Expected 5 columns on line " << (ValidLineCount + 1) << endl;
```
fhagemann:
```suggestion
        if (g_Verbosity >= c_Error) {
          cout << m_XmlTag << " ERROR in LoadSimCCEFile: Expected 5 columns on line " << (ValidLineCount + 1) << endl;
        }
```

include/MModuleTrappingCorrection.h
```diff
@@ -0,0 +1,163 @@
...
+
+
+  // private methods
+  private:
+
+
+  // protected members:
+ protected:
+
+  unordered_map<int, vector<double>> m_SimCCE;
```
fhagemann: This variable `m_SimCCE` seems to be unused:
```suggestion
```

include/MModuleTrappingCorrection.h
```diff
@@ -0,0 +1,163 @@
...
+  double m_SimCCE_Energy;
+  MString m_SimCCEFile;
```
fhagemann: Consider renaming this
```suggestion
  MString m_SimCCEFileName;
```
here and elsewhere

include/MModuleTrappingCorrection.h
```diff
@@ -0,0 +1,163 @@
...
+ 
+  // unordered_map<int, MDDetector*> m_Detectors;
+  vector<unsigned int> m_DetectorIDs;
```
fhagemann: What is this used for (other than checking if it is empty)? Can this also go?
```suggestion
```

fhagemann (commented):
Here is my feedback from just looking at the code. 
I will also try to compile and run both the module and the app to provide more feedback on that as well! :)

fhagemann:
The example file in `resource/dee` and the results after applying the inverse charge trapping in the DEE looks reasonable:

image
image
image
image
image

src/MModuleTrappingCorrection.cxx
sophieehaight: Yes, that's the case atm. There was conversation recently about this but I don't remember the consensus.

apps/TrappingCorrectionCs137.cxx
sophieehaight: yes I suppose :)

src/MGUIOptionsLoaderMeasurementsHDF.cxx
sophieehaight: oh no this looks like a mistake and I got the GUI options files mixed up when editing. I will add these lines back and push the corrected file in my next commit.

src/MModuleTrappingCorrection.cxx
```diff
@@ -387,7 +370,9 @@ bool MModuleTrappingCorrection::LoadSimCCEFile(MString FileName)
         
         ValidLineCount++;
       } else {
-        cout << "ERROR in LoadSimCCEFile: Expected 5 columns on line " << (ValidLineCount + 1) << endl;
+        if (g_Verbosity >= c_Error) {
+          cout << "ERROR in LoadSimCCEFile: Expected 5 columns on line " << (ValidLineCount + 1) << endl;
+        }
```
fhagemann: Should this also close the file (`SimCCEFile.Close()`) and return `false`?

src/MGUIOptionsLoaderMeasurementsHDF.cxx
```diff
@@ -138,7 +137,6 @@ bool MGUIOptionsLoaderMeasurementsHDF::OnApply()
   dynamic_cast<MModuleLoaderMeasurementsHDF*>(m_Module)->SetFileName(m_FileSelectorHDF->GetFileName());
   dynamic_cast<MModuleLoaderMeasurementsHDF*>(m_Module)->SetLoadContinuationFiles(m_LoadContinuationFiles->IsOn());
   dynamic_cast<MModuleLoaderMeasurementsHDF*>(m_Module)->SetFileNameStripMap(m_FileSelectorStripMap->GetFileName());
-  dynamic_cast<MModuleLoaderMeasurementsHDF*>(m_Module)->SetIncludeNearestNeighbor(m_IncludeNearestNeighbor->IsOn());
```
fhagemann: You might also want to reintroduce this thing here ;)

sophieehaight: So odd! I don't recall changing this. I'll add this back to have it synched up with the version on develop/em

src/MModuleTrappingCorrection.cxx
sophieehaight: addressed in [recent commit](https://github.com/cositools/nuclearizer/pull/184#:~:text=deleted%20unnecessary%20output-,be838f7,-codecov%2Dcommenter)

src/MModuleTrappingCorrection.cxx
sophieehaight: addressed in [this](https://github.com/cositools/nuclearizer/pull/184#:~:text=deleted%20unnecessary%20output-,be838f7,-codecov%2Dcommenter) commit

src/MModuleTrappingCorrection.cxx
sophieehaight: addressed in [this](https://github.com/cositools/nuclearizer/pull/184#:~:text=deleted%20unnecessary%20output-,be838f7,-codecov%2Dcommenter) commit

src/MModuleTrappingCorrection.cxx
sophieehaight: I addressed this in [this commit](https://github.com/cositools/nuclearizer/pull/184#:~:text=deleted%20unnecessary%20output-,be838f7,-codecov%2Dcommenter). For the XmlTag I was following the convention used in the depth calibration but I'm also not sure if it's used in other modules. 

Tbh when my verbosity is set to info it overflows the command line and crashes nuclearizer :/. I think there's too much info outputted atm from the depth calibration.

include/MModuleTrappingCorrection.h
sophieehaight: I think I want to hang on to the error message associated with this variable until I get a chance to test the module on data from multiple detectors at once. If it works well then I think it's okay to remove the error message on lines 134-139

fhagemann: Then maybe add a `// TODO` comment prior to lines 134-139 so that we remember :)

src/MGUIOptionsLoaderMeasurementsHDF.cxx
sophieehaight: fixed [here]!(https://github.com/cositools/nuclearizer/pull/184/commits/c19ab79a17d83f09ebbc12c9c0ff78a9d0d5e7a5)

src/MModuleTrappingCorrection.cxx
sophieehaight: fixed in https://github.com/cositools/nuclearizer/pull/184/commits/43116c991f973af2c4cb60624525cd9e709d64c4

include/MModuleTrappingCorrection.h
sophieehaight: I updated it here: https://github.com/cositools/nuclearizer/pull/184/commits/e55d0643bcebc84fdc7e59027ba107af40a5cbcf

include/MModuleTrappingCorrection.h
sophieehaight: I removed it in this commit: https://github.com/cositools/nuclearizer/pull/184/commits/e55d0643bcebc84fdc7e59027ba107af40a5cbcf

apps/TrappingCorrectionCs137.cxx
sophieehaight: If you run "TrappingCorrectionCs137 -h" it will tell you the app inputs to run it from the command line

include/MHit.h
sophieehaight: updated in: https://github.com/cositools/nuclearizer/pull/184/commits/0513b47af46d61b0946fe42d704265224c437824

resource/dee/dummy_trapping_parameters_singledetector.csv
sophieehaight: So My plan for multiple detectors is to have commented line with the detector ID above the parameters and CCEs. e.g:

# detector ID
<detector ID>
# A_HV,A_LV,B,C
<trapping params>
# z_depth_cm,e_CCE_HV,h_CCE_HV,e_CCE_LV,h_CCE_LV
<sim CCE>

Is this in line with how you were going to write the param files for the depth calibration?

apps/TrappingCorrectionCs137.cxx
sophieehaight: The txt files give the lineshift across depth but they need to be fit to sim CCE curves in python to get the trapping param files so I'm not sure how helpful these will be until someone else needs to learn to produce trapping params. Just probably not something we need incorporated in nuclearizer atm.

apps/TrappingCorrectionAm241.cxx
sophieehaight: removed in https://github.com/cositools/nuclearizer/pull/184/commits/d8e1b98f47a3d550f06297d437252464ce9592d9 !

apps/TrappingCorrectionCs137.cxx
sophieehaight: updated in https://github.com/cositools/nuclearizer/pull/184/commits/675aab7c40c9169e8da8fdb3c1d00f8faeeac94b

resource/dee/dummy_trapping_parameters_singledetector.csv
sophieehaight: I updated the param file and the CCE read in function in commit https://github.com/cositools/nuclearizer/pull/184/commits/87e4ae84fef950608bb728c53bdc07774b7ae80f

apps/TrappingCorrectionCs137.cxx
fhagemann: Ah sorry, I was refering to a `txt` file used to pass multiple filenames to the app.
Would that just be
```
gse_filename1.hdf5
gse_filename2.hdf5
gse_filename3.hdf5
...
```
?

resource/dee/dummy_trapping_parameters_singledetector.csv
fhagemann: Great! My final question would be: what about `detector ID` to concatenate files?

apps/TrappingCorrectionCs137.cxx
sophieehaight: oh yeah that's it

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+{
+  // Construct an instance of MModuleTrappingCorrection
+
+  // Set all module relevant information
+
+  // Set the module name --- has to be unique
+  m_Name = "Trapping Correction"; // - correcting energies for charge trapping (by Sophie);
+
+  // Set the XML tag --- has to be unique --- no spaces allowed
+  m_XmlTag = "TrappingCorrection";
```
fhagemann: Let's be consistent with XML tag naming here (see #193)
```suggestion
  m_XmlTag = "XmlTagTrappingCorrection";
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+  m_DepthCalibration = (MModuleDepthCalibration*) S->GetAvailableModuleByXmlTag("DepthCalibration");
+  if (m_DepthCalibration == nullptr) {
+    if (g_Verbosity >= c_Error) {
+      cout << "ERROR in MModuleTrappingCorrection::Initialize: couldn't resolve pointer to Depth Calibration Module... need access to this module for depth resolution lookup!" << endl;
+    }
+       return false;
+  }
+
+// TO DO: remove this check once we successfully process multiple detectors
+  m_DetectorIDs = m_DepthCalibration-> GetDetectorIDs();
```
fhagemann:
```suggestion
  m_DetectorIDs = m_DepthCalibration->GetDetectorIDs();
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+
+  if (m_DetectorIDs.empty()) {
+    if (g_Verbosity >= c_Error) {
+      cout << "ERROR in MModuleTrappingCorrection::Initialize: Depth Calibration has no registered detector IDs!" << endl;
+    }
+      return false;
+  }
```
fhagemann:
```suggestion
  if (m_DetectorIDs.empty()) {
    if (g_Verbosity >= c_Error) {
      cout << "ERROR in MModuleTrappingCorrection::Initialize: Depth Calibration has no registered detector IDs!" << endl;
    }
    return false;
  }
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+/////////////////////////////////////////////////////////////////////////////////
+
+void MModuleTrappingCorrection::Finalize()
+{
+  MModule::Finalize();
+
+  if (m_ExpoSpectrum == nullptr) {
+    cout << "ERROR in MModuleTrappingCorrection::Finalize: Expo plot spectrum is null." << endl;
+    return;
+  }
```
fhagemann:
```suggestion
  if (m_ExpoSpectrum == nullptr) {
    if (g_Verbosity >= c_Error) {
      cout << "ERROR in MModuleTrappingCorrection::Finalize: Expo plot spectrum is null." << endl;
    }
    return;
  }
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+    fitFunc->SetParameter("Amplitude", hist->GetBinContent(hist->GetMaximumBin()));
+    hist->Fit(fitFunc, "RQ");
+
+    double mu   = fitFunc->GetParameter("x0 (Mu)");
+    double fwhm = 2.35482 * fitFunc->GetParameter("Sigma Gauss");
+
+    cout << "\n" << m_XmlTag << " --- " << titleLabel << " ---" << endl;
+    cout << "  Centroid (Mu)        : " << mu << " keV" << endl;
+    cout << "  Fitted Gaussian FWHM : " << fwhm << " keV" << endl;
+    cout << "  Direct Histogram FWHM: " << directFWHM << " keV" << endl;
```
fhagemann: Put this in `c_Info`?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+  cout << "\n========================================================" << endl;
+  cout << "                 SUMMARY COMPARISON                     " << endl;
+  cout << "========================================================" << endl;
+  cout << " LV Side: " << endl;
+  cout << "   Raw Centroid: " << lv_raw.first << " keV | Corrected Centroid: " << lv_corr.first << " keV" << endl;
+  cout << "   Raw FWHM    : " << lv_raw.second << " keV | Corrected FWHM    : " << lv_corr.second << " keV" << endl;
+  cout << " HV Side: " << endl;
+  cout << "   Raw Centroid: " << hv_raw.first << " keV | Corrected Centroid: " << hv_corr.first << " keV" << endl;
+  cout << "   Raw FWHM    : " << hv_raw.second << " keV | Corrected FWHM    : " << hv_corr.second << " keV" << endl;
+  cout << "========================================================\n" << endl;
```
fhagemann: `c_Info`?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,614 @@
...
+    cout << "  Fitted Gaussian FWHM : " << fwhm << " keV" << endl;
+    cout << "  Direct Histogram FWHM: " << directFWHM << " keV" << endl;
+
+    delete fitFunc;
+    return std::make_pair(mu, fwhm);
+  };
+
+  cout << "\n========================================================" << endl;
+  cout << "      TRAPPING CORRECTION SPECTRUM FIT RESULTS          " << endl;
+  cout << "========================================================" << endl;
```
fhagemann: `c_Info`?

fhagemann (commented):
Final code-style review, will give it a run to see that it compiles.
But then this should be good to go (tackling reading trapping parameters for multiple detectors in a follow-up PR)

src/MModuleTrappingCorrection.cxx
sophieehaight: I use this output to evaluate how well the trapping correction is working. If I change the verbosity to info, the comand line becomes overloaded (I believe w info from the depth cal) and nuclearizer crashes. I agree this info doesn't need to be repeated so I can change the verbosity for the following comments but I do need it read out somewhere. 

I think we probably need to clean up the outputs for the info verbosity right now because right now it's outputting so much info that I can't receive the info I'm trying to get from the trapping correction.

src/MModuleTrappingCorrection.cxx
sophieehaight: I removed these lines because they output the same info as lines 277-280

src/MModuleTrappingCorrection.cxx
sophieehaight: I removed this output

src/MModuleTrappingCorrection.cxx
fhagemann: @zoglauer maybe we can introduce the chatty verbosity from megalib also to nuclearizer, and move some of the info messages to chatty?

sophieehaight: yeah that could be a good idea!

fhagemann (approved):
This would be good from my side. Just gathering here the open TODOs (for follow-up PRs):
- [ ] Allow for a charge trapping parameters file for multiple detectors, at not just one
- [ ] Think about flushing output to `c_Info` without overloading the output
- [ ] Maybe find an alternative to `Interpolate` to avoid code duplication (maybe from an existing package or another module --> the depth calibration also used some sort of interpolation algorithm to interpolate the CTD-to-depth spline).

src/MModuleTrappingCorrection.cxx
sophieehaight: I'm actually not seeing this convention in the depth or energy calibration modules? It looks like the XML tag is just the name of the module

sophieehaight: Once I change the name of the xml tag, nuclearizer can no longer recognize the module. Is there somewhere else in assembly I need to update the xml tag? I was looking but I couldn't find anything

fhagemann: This will affect reading in a nuclearizer config file (`*.cfg`). In those, you might need to update all occurences of `TrappingCorrection` with `XmlTrappingCorrection`.

fhagemann: So either update the `.nuclearizer.cfg` in your home directory, or the respective `.cfg` that you load when running nuclearizer.

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+
+
+/////////////////////////////////////////////////////////////////////////////////
+
+
+bool MModuleTrappingCorrection::AnalyzeEvent(MReadOutAssembly* Event) 
+{
+
+  if (Event->GetGuardRingVeto() || m_DetectorParamMap.empty()) {
+  }
```
fhagemann: This does nothing right now?
Should this be returning true? Or be part of the loop and skipping over those with `continue`?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+  m_DepthCalibration = (MModuleDepthCalibration*) S->GetAvailableModuleByXmlTag("DepthCalibration");
+  if (m_DepthCalibration == nullptr) {
+    if (g_Verbosity >= c_Error) {
+      cout << "ERROR in MModuleTrappingCorrection::Initialize: couldn't resolve pointer to Depth Calibration Module... need access to this module for depth resolution lookup!" << endl;
+    }
+       return false;
+  }
+
+// TO DO: remove this check once we successfully process multiple detectors
+  m_DetectorIDs = m_DepthCalibration->GetDetectorIDs();
```
fhagemann: Can we successfully process multiple detectors now? 😅

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+        bool is_depth_out_of_bounds = (depth_val < min_param_depth || depth_val > max_param_depth);
+
+        double rawLVEnergy = LVSH->GetEnergy(); 
+
+        if (HasExpos() == true) {
+          m_ExpoSpectrum->AddEnergyInitial(rawLVEnergy, LVSH->IsNearestNeighbor(), LVSH->IsLowVoltageStrip());
+        }
+
+        double correctedLVEnergy = rawLVEnergy;
+        if (!is_depth_out_of_bounds) {
```
fhagemann:
```suggestion
        if (is_depth_out_of_bounds == false) {
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+        double max_param_depth = std::max(detData.m_Depths.front(), detData.m_Depths.back());
+        bool is_depth_out_of_bounds = (depth_val < min_param_depth || depth_val > max_param_depth);
+
+        double rawHVEnergy = HVSH->GetEnergy(); 
+        
+        if (HasExpos() == true) {
+          m_ExpoSpectrum->AddEnergyInitial(rawHVEnergy, HVSH->IsNearestNeighbor(), HVSH->IsLowVoltageStrip());
+        }
+        double correctedHVEnergy = rawHVEnergy;
+        if (!is_depth_out_of_bounds) {
```
fhagemann:
```suggestion
        if (is_depth_out_of_bounds == false) {
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+    TF1* fitFunc = GeneratePhotopeakFunction();
+    fitFunc->SetParameter("Amplitude", hist->GetBinContent(hist->GetMaximumBin()));
+    hist->Fit(fitFunc, "RQ");
+
+    double mu        = fitFunc->GetParameter("x0 (Mu)");
+    double gaussFWHM = 2.35482 * fitFunc->GetParameter("Sigma Gauss");
+    
+    // Evaluate full function FWHM over the fit window (645 keV to 675 keV)
+    double xMinFit = 645.0;
+    double xMaxFit = 675.0;
```
fhagemann: This seems to be tailored to Cs137. Is this what we want? What if we pass data from other isotopes through `MModuleTrappingCorrection`?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+    cout << "=======================================================" << endl;
+    cout << " Centroid Shift (Delta Mu)     : " << lineShift << " keV (" << mu_raw << " -> " << mu_corr << ")" << endl;
+    cout << " Gaussian FWHM Change          : " << deltaGaussFWHM << " keV (" << gauss_fwhm_raw << " -> " << gauss_fwhm_corr << ")" << endl;
+    cout << " Full Fit Function FWHM Change : " << deltaFullFitFWHM << " keV (" << full_fwhm_raw << " -> " << full_fwhm_corr << ")" << endl;
+    cout << "=======================================================\n" << endl;
+  };
+
+  // --- OUTPUT DELTA COMPARISONS ---
+  PrintTrappingCorrectionSummary("LOW VOLTAGE (LV) STRIPS", lv_raw, lv_corr);
+  PrintTrappingCorrectionSummary("HIGH VOLTAGE (HV) STRIPS", hv_raw, hv_corr);
```
fhagemann: This (and maybe even the code before determining the fits) should be enclosed in `g_Verbosity >= c_Info`

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+  
+  return; 
+}
+/////////////////////////////////////////////////////////////////////////////////
+
+
+bool MModuleTrappingCorrection::LoadSimCCEFile(MString FileName)
+{
+  MFile SimCCEFile;
+  if (!SimCCEFile.Open(FileName)) return false;
```
fhagemann: Add an error message here:
```suggestion
  if (SimCCEFile.Open(FileName) == false) {
    if (g_Verbosity >= c_Error) {
      cout << m_XmlTag << ": ERROR: Could not open CCE file " << FileName << endl;
    }
    return false;
  }
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+
+  m_DetectorParamMap.clear();
+  MString Line;
+  int currentDetID = -1;
+
+  while (SimCCEFile.ReadLine(Line)) {
+    Line = Line.Strip();
+    if (Line.IsEmpty()) continue;
```
fhagemann:
```suggestion
    if (Line.IsEmpty() == true) continue;
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+      det.m_Depths.push_back(Tokens[0].Strip().ToDouble());
+      det.m_CCEs_HV_e.push_back(Tokens[1].Strip().ToDouble());
+      det.m_CCEs_HV_h.push_back(Tokens[2].Strip().ToDouble());
+      det.m_CCEs_LV_e.push_back(Tokens[3].Strip().ToDouble());
+      det.m_CCEs_LV_h.push_back(Tokens[4].Strip().ToDouble());
+    }
+  }
+
+  SimCCEFile.Close();
+  return !m_DetectorParamMap.empty();
```
fhagemann: Also add an error message here:
```suggestion

  if (m_DetectorParamMap.empty() == true) {
    if (g_Verbosity >= c_Error) {
      cout << m_XmlTag << ": ERROR: Obtained an empty charge trapping parameter map when reading CCE file " << FileName << endl;
    }
    return false;
  }
  
  return true;
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+
+    // Read parameters (A_HV, A_LV, B, C)
+    if (det.m_Depths.empty() && Tokens.size() == 4) {
+      det.m_ParamA_HV = Tokens[0].Strip().ToDouble();
+      det.m_ParamA_LV = Tokens[1].Strip().ToDouble();
+      det.m_ParamB    = Tokens[2].Strip().ToDouble();
+      det.m_ParamC    = Tokens[3].Strip().ToDouble();
+    }
+    // Read CCE depth curves
+    else if (Tokens.size() == 5 && !Line.Contains("z_depth")) {
```
fhagemann:
```suggestion
    else if (Tokens.size() == 5 && Line.Contains("z_depth") == false) {
```
but shouldn't this be in a line starting with `#`, so it's already covered?

Then, you might just want to got for:
```suggestion
    else if (Tokens.size() == 5) {
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+    if (Line.IsEmpty()) continue;
+
+    // Detect section header: "### <Detector ID>"
+    if (Line.BeginsWith("###")) {
+      std::vector<MString> Tokens = Line.Tokenize(" ");
+      if (Tokens.size() >= 1) {
+        currentDetID = Tokens.back().Strip().ToInt();
+      }
+      continue;
+    }
```
fhagemann: How do you deal with the case, if one detector ID appears multiple times in the file, let's say:
```
### 0
[...]

### 0
[...]
```
Should this throw an error?

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+
+  SimCCEFile.Close();
+  return !m_DetectorParamMap.empty();
+}
+
+/////////////////////////////////////////////////////////////////////////////////
+
+double MModuleTrappingCorrection::GetSimBasedCorrectedEnergy(double depth_val, double uncorrected_energy, const std::vector<double>& depths, const std::vector<double>& sim_cce_sorted_e, const std::vector<double>& sim_cce_sorted_h, double paramB, double paramC)
+{
+  if (sim_cce_sorted_e.empty() || sim_cce_sorted_h.empty()) {
```
fhagemann:
```suggestion
  if (sim_cce_sorted_e.empty() == true || sim_cce_sorted_h.empty() == true) {
```

src/MModuleTrappingCorrection.cxx
```diff
@@ -0,0 +1,603 @@
...
+/////////////////////////////////////////////////////////////////////////////////
+
+
+bool MModuleTrappingCorrection::ReadXmlConfiguration(MXmlNode* Node)
+{
+  //! Read the configuration data from an XML node
+
+  MXmlNode* SimCCEFileNameNode = Node->GetNode("SimCCEFileName");
+  if (SimCCEFileNameNode != nullptr) {
+  m_SimCCEFileName = SimCCEFileNameNode->GetValue();
```
fhagemann:
```suggestion
    m_SimCCEFileName = SimCCEFileNameNode->GetValue();
```

fhagemann (commented):
Some code-style and some more major comments.

One general comment:
you are using a lot of snake-case when assigning variable names, we might want to use camel-case consistently here (e.g. `SimCCESortedHoles` instead of `sim_cce_sorted_h`)

src/MModuleTrappingCorrection.cxx
sophieehaight: addressed: [a9e7a30](https://github.com/cositools/nuclearizer/pull/184/commits/a9e7a30af86dd2a67998aa8c1a0392988539ecab)

src/MModuleTrappingCorrection.cxx
sophieehaight: addressed: [a9e7a30](https://github.com/cositools/nuclearizer/pull/184/commits/a9e7a30af86dd2a67998aa8c1a0392988539ecab)

src/MModuleTrappingCorrection.cxx
sophieehaight: addressed: [a9e7a30](https://github.com/cositools/nuclearizer/pull/184/commits/a9e7a30af86dd2a67998aa8c1a0392988539ecab)

src/MModuleTrappingCorrection.cxx
sophieehaight: yes, good catch. we should be skipping these events. I fixed this here: [f58efca](https://github.com/cositools/nuclearizer/pull/184/commits/f58efca84d9d23d8f1c7de07cf6e7afe4e6725cb)

src/MModuleTrappingCorrection.cxx
sophieehaight: yep! all ready now :) ([4e6a44f](https://github.com/cositools/nuclearizer/pull/184/commits/4e6a44fea86e04db0226b497be078c14a38dc636))

src/MModuleTrappingCorrection.cxx
sophieehaight: fixed: [041e8ed](https://github.com/cositools/nuclearizer/pull/184/commits/041e8edf6f2746cdf18594ef8a1f144a7bf6a117)

src/MModuleTrappingCorrection.cxx
sophieehaight: fixed: [041e8ed](https://github.com/cositools/nuclearizer/pull/184/commits/041e8edf6f2746cdf18594ef8a1f144a7bf6a117)

src/MModuleTrappingCorrection.cxx
sophieehaight: fixed: [041e8ed](https://github.com/cositools/nuclearizer/pull/184/commits/041e8edf6f2746cdf18594ef8a1f144a7bf6a117)

src/MModuleTrappingCorrection.cxx
sophieehaight: fixed here: [88cf99f](https://github.com/cositools/nuclearizer/pull/184/commits/88cf99f08cf96e1adacac1c4a9edea30c5ded4c7)

src/MModuleTrappingCorrection.cxx
sophieehaight: jk it needed to return false s.t the analysis of the event is stopped. This is updated here: [9a18863](https://github.com/cositools/nuclearizer/pull/184/commits/9a188634326420552c335eba1075b0d55cbbeeda)

src/MModuleTrappingCorrection.cxx
sophieehaight: this is adjusted here and I fixed the delta FWHM calculations as well: [7b4cdbf](https://github.com/cositools/nuclearizer/pull/184/commits/7b4cdbfe53a9ad1dcf8aa890b4638c6fd33b9339)

src/MModuleTrappingCorrection.cxx
sophieehaight: I added a check for this here: [3b30c45](https://github.com/cositools/nuclearizer/pull/184/commits/3b30c45584623f85a416e14622dc9176797b2ff2)

src/MModuleTrappingCorrection.cxx
sophieehaight: I set it up in commits [8f30f9b](https://github.com/cositools/nuclearizer/pull/184/commits/8f30f9b65640ee477bb32656e0bea877a05257c2), [b95dc1e](https://github.com/cositools/nuclearizer/pull/184/commits/b95dc1eec7074f09e974f04ad0221f6ebfa26019), and [f191838](https://github.com/cositools/nuclearizer/pull/184/commits/f19183869200578cf966abad85f5c57aa971a759) (took a couple tries to get my axes to update correctly here). I let the user define the energy of the photopeak they are tryting to fit on the nuclearizer expo. The trapping correction module then takes that input and uses it to define the Mu and min and max energies for the photopeak fits measuring resolutions pre and post trapping correction. I kept the default energy for Cs-137 as a benchmark.

include/MGUIOptionsTrappingCorrection.h
```diff
@@ -0,0 +1,87 @@
...
+#include "MModule.h"
+
+
+// Forward declarations:
+
+
+////////////////////////////////////////////////////////////////////////////////
+
+
+//! The user interface for the universal energy calibration
```
ckierans: `//! The user interface for the trapping correction`

include/MGUIOptionsTrappingCorrection.h
```diff
@@ -0,0 +1,87 @@
+/*
+ * MGUIOptionsTrappingCorrection.h
+ *
+ * Copyright (C) by Andreas Zoglauer.
```
ckierans: Add you name :)

src/MModuleTrappingCorrection.cxx
```diff
@@ -615,7 +610,7 @@ TF1* MModuleTrappingCorrection::GeneratePhotopeakFunction()
 
   // Set boundary limits to stabilize convergence
   PhotopeakFunction->SetParLimits(0, 1, 1e8);
-  PhotopeakFunction->SetParLimits(1, fitMin, fitMax);     // Keeps peak centered around Mu 
+  PhotopeakFunction->SetParLimits(1, 645, 675);     // Keeps peak centered around 662 keV
```
fhagemann: Your last pushes reverted back to the fixed window around 662keV.
Is this intended?

fhagemann:
I've lost track. Is this ready for a final review?

resource/dee/dummy_payload_trapping_parameters.csv
fhagemann: Do you also plan to add the individual detector files to `resource/calibration/COSISMEX/unit`?

src/MModuleTrappingCorrection.cxx
sophieehaight: my mistake! I was testing the fit with and without the dynamic fitting window. This is fixed in [6321870](https://github.com/cositools/nuclearizer/pull/184/commits/6321870f09e5c6e7ea8750e8b914a79ec48e9d59)

resource/dee/dummy_payload_trapping_parameters.csv
sophieehaight: yes, I will do this today

PR #188, opened 2026-08-11
CHG: Harden MStripMap and add unit test

zoglauer:
Claude and I hardened MStripMap and added a unit test.

However, one test is failing, and Claude convinced me that there is the potential for a real bug in the code. However, I am not familiar enough with the strip map parsing, so I cannot tell for sure. Here is Claude's question to @fhagemann .

⚠️ Known failing test — open question for @fhagemann

UTNStripMap reports 220 passed / 1 failed. The failure is intentional and marked KNOWN FAILURE - open question in the source. Everything else is green, including both end-to-end tests and UTNModuleLoaderMeasurementsHDF.

What it records. A (detector, side, strip) tuple must map to exactly one read-out channel, so a polarity update that collides should be rejected. When I made it an error, both committed datasets stopped loading.

Why. The config JSON in both datasets reports:

detector 0:      primary SP = 1,1,1     secondary SP = 0,0,0
detectors 1-15:  primary SP = 1,1,1     secondary SP = 1,1,1

Both files are identical here — 3 × SP:0, 93 × SP:1 — despite being recorded four months apart (gse_20251016T191639, gse_20260217T124025). Meanwhile the strip maps assign polarity consistently across all 2080 rows of each:

is_primary = 1  ->  side 0 (LV)    1040 rows
is_primary = 0  ->  side 1 (HV)    1040 rows

For detector 0 the update is a no-op. For detectors 1-15 it flips every secondary ASIC onto the LV side, on top of the primary strips already there — 975 collisions, exactly 15 detectors × 65 strips. Taken at face value, both physical sides of 15 of our 16 detectors are labelled low voltage.

This is not specific to these two files: the output posted on #100 (5 March, FM assembly at SSL) shows the same shape — detector 0 with differing sides, detectors 1-15 identical on both.

Ruled out: inverting the SP → LV/HV mapping does not help. The collision comes from primary and secondary carrying the same value, and a global relabel preserves equality — I tried it, identical failure, and it additionally puts detector 0 at odds with its strip map.

The question. On #100 the design was never settled: the request was to switch between strip maps based on the JSON ASIC setup, replicating Clio's Python; the implementation instead overrides the side flag in a single map, which is what UpdateASICPolarities() does today. So:

1. Was the override approach a deliberate replacement for the switching design, or a working assumption? With this data it is the difference between a no-op and mislabelling 15 detectors.
2. Was the suspected bug in the Python reference ever tracked down? Does the Python treat SP in the secondary section as a polarity in its own right, or derive the secondary side as the complement of the primary? If derived, everything becomes consistent and the update is a no-op on all our data.
3. Is SP meant to be trustworthy per-ASIC, or only for detector 0 in these files?

Meanwhile, UpdateASICPolarities() keeps the last colliding entry and returns true — the behaviour the committed references were generated with — so the loader and both end-to-end tests keep passing. A TODO: marks where to restore the rejection. The reference .tra files encode detectors 1-15 as all-LV and are not being regenerated until this is answered.

🤖 Generated with Claude Code

### DIFF
--- a/CodingConventions.md
+++ b/CodingConventions.md
@@ -187,6 +187,16 @@ X, DataPoint, IsNonZero
   ```
 
 
+For a simple one-statement diagnostic guard, prefer the compact form used
+throughout the project:
+
+```cpp
+if (g_Verbosity >= c_Error) cout << "Error: Unable to open file" << endl;
+```
+
+Do not expand this form to a multi-line braced block unless the guarded code
+contains multiple statements or the longer form materially improves clarity.
+
 ### 6. **Whitespace in Expressions**
 
 - **Space Around Operators**: Always add a space before and after most binary operators (=, +, -, &&, etc.). The exce[tion are * and / in math expresions, to visualize the order of operations.
--- a/include/MModuleLoaderMeasurementsHDF.h
+++ b/include/MModuleLoaderMeasurementsHDF.h
@@ -299,6 +299,10 @@ class MModuleLoaderMeasurementsHDF : public MModuleLoaderMeasurements
   //! The map which ASICs are HV/LV for HDFv2
   vector<map<bool, vector<bool>>> m_ASICPolarities;
 
+  //! The detector IDs which were enabled during the run, read from the config JSON
+  //! Empty if the config JSON does not contain the enabled flags
+  vector<unsigned int> m_EnabledDetectors;
+
   //! The strip map
   MStripMap m_StripMap;
   
--- a/include/MStripMap.h
+++ b/include/MStripMap.h
@@ -17,6 +17,7 @@
 
 
 // Standard libs:
+#include <map>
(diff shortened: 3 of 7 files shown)


fhagemann (approved):
I just had a deeper look at this:
- the regex JSON reading code looks good (I checked with my Julia implementation and it gives the expected results)
- the way it determines `DetectorIsEnabled` and `m_EnabledDetectors` also looks fine
- `RestrictToEnabledDetectors` also looks fine
- Returning `false` in `UpdateASICPolarities` and not updating polarities when creating duplicate entries is also fine.

The newest code might only break if the detector settings in the JSON config are not saved in the order of ascending detector IDs, i.e. if the first entry in the JSON config does not belong to detector ID 0, etc. -- but I don't think that this is ever the case or something to worry about.

PR #191, opened 2026-08-12
CHG: MAssembly hardening and unit tests

zoglauer:
Unit test, bug fixes, and hardening for MAssembly class

### DIFF
--- a/unittests/UTNAssembly.cxx
+++ b/unittests/UTNAssembly.cxx
@@ -1,0 +1,792 @@
+/*
+ * UTNAssembly.cxx
+ *
+ * Copyright (C) by Andreas Zoglauer.
+ * All rights reserved.
+ *
+ * Please see the source-file for the copyright-notice.
+ *
+ */
+
+
+// Standard libs:
+#include <cstring>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <streambuf>
+#include <vector>
+using namespace std;
+
+// ROOT libs:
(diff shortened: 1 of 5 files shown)


unittests/UTNAssembly.cxx
```diff
@@ -0,0 +1,792 @@
...
+  MAssembly Assembly;
+  MSupervisor* Supervisor = MSupervisor::GetSupervisor();
+
+  // The supervisor is a process-wide singleton whose available-module list only ever grows,
+  // so every check here is a lookup by XML tag - never a total count and never an assertion
+  // that a module occurs only once
+  vector<MString> XmlTags = { "XmlTagLoaderSimulations", "XmlTagMeasurementLoaderROA", "XmlTagMeasurementLoaderHDF",
+                              "XmlTagMeasurementLoaderFITS", "XmlTagMeasurementLoaderL0", "XmlTagDEESMEX",
+                              "XmlTagEventFilter", "EnergyCalibration", "XmlTagStripPairingMultiRoundChiSquare",
+                              "XmlTagStripPairingChiSquare", "DepthCalibration", "XmlTagEventSaver",
```
fhagemann: Looks like `EnergyCalibration` and `DepthCalibration` are the only two modules that do not start their `XmlTag` with `XmlTag`. Should this be updated? Maybe also make this a unit test to check that an `XmlTag` starts with `XmlTag` to prevent this from happening for future modules?

Issue #63, opened 2025-10-24
Refine TACcut coincidence window


NicoleRodriguezCavero:
MModuleTACcut.cxx: Event coincidence window subject to change pending more analysis.

PR #194, opened 2026-08-14
Separate TAC calibration and cuts

### LINKED ISSUE #64: Refine TACcut TotalOffset
MModuleTACcut.cxx: Currently the TotalOffset value is hard coded. Should compare with other electronics timing contributions for a more precise value.

### LINKED ISSUE #193: Consistent use of XML tags
The unit tests for `MAssembly` revealed some inconsistencies in how we assign XML tags to nuclearizer modules:
https://github.com/zoglauer/nuclearizer/blob/16400db68d2d3982864b5c606854047ad415408a/unittests/UTNAssembly.cxx#L259-L265
```
  vector<MString> XmlTags = { "XmlTagLoaderSimulations", "XmlTagMeasurementLoaderROA", "XmlTagMeasurementLoaderHDF",
                              "XmlTagMeasurementLoaderFITS", "XmlTagMeasurementLoaderL0", "XmlTagDEESMEX",
                              "XmlTagEventFilter", "EnergyCalibration", "XmlTagStripPairingMultiRoundChiSquare",
                              "XmlTagStripPairingChiSquare", "DepthCalibration", "XmlTagEventSaver",
                              "XmlTagSaverMeasurementsL0", "XmlTagSaverMeasurementsFITS", "XmlTagTransmitterRealta",
                              "XmlTagResponseGenerator", "XmlTagRevan", "XmlTagTACcut", "XmlTagDiagnostics",
                              "XmlTagDiagnosticsEnergyPerStrip" };
```
While most of the XML tags start with `XmlTag`, there is some modules where the naming is a bit awkward:
- `EnergyCalibration` (instead of `XmlTagEnergyCalibration`)
- `DepthCalibration` (instead of `XmlTagDepthCalibration`)

In that sense: 
the TAC calibration and cut module right now is called `MModuleTACcut`, even though it's essentially doing the same thing as `EnergyCalibration` (calibrating ADC values, and then applying cuts/thresholds). 
If we are to rename XML tags, which might be a breaking change for people using nuclearizer config files, we should also think about renaming the module/XML tag to `MModuleTACCalibration` and `XmlTagTACCalibration` instead.
(and maybe split the code doing TAC calibration and TAC cuts rather than having everything in the big `AnalyzeEvent` function).

NicoleRodriguezCavero:
Currently, `MModuleTACCut` performs the TAC calibration and applied the TAC cuts all within the `AnalyzeEvent()` functions. This PR:
- Separates the `ApplyTACCal()` and `ApplyTACCuts()` functions and calls them in `AnalyzeEvent()`
- No longer requires a taccut file and instead has a GUI option to set the coincidence window used in the TAC cuts
- All non-GR events regardless of whether they have slow or fast timing are now calibrated to ns
- Renames module from `MModuleTACCut` to `MModuleTACCalibration`
- Renames XMLtags from `TACcut` to `TACcalibration` addressing part of #193


src/MModuleTACcal.cxx
```diff
@@ -0,0 +1,496 @@
...
+  return true;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+
+void MModuleTACcal::Finalize()
+{
+  MModule::Finalize();
+}
```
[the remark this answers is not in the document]
NicoleRodriguezCavero: Sounds good, I'll also go ahead and clear `m_DetectorIDs`.

src/MModuleTACcal.cxx
```diff
@@ -0,0 +1,496 @@
+/*
+ * MModuleTACcal.cxx
```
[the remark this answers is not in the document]
NicoleRodriguezCavero: This is a good point and I shall bring it up tomorrow during out meeting. If that is the case I can make some quick changes :)

include/MAssembly.h
```diff
@@ -53,26 +53,27 @@ class MAssembly
...
-  static const uint64_t c_NoRestriction            = (1 << 17);
-  static const uint64_t c_EventSaver               = (1 << 18);
-  static const uint64_t c_EventTransmitter         = (1 << 19);
-  static const uint64_t c_PositionDetermiation     = (1 << 20);
-  static const uint64_t c_Statistics               = (1 << 21);
-  static const uint64_t c_FlagHits                 = (1 << 22);
-  static const uint64_t c_Diagnostics              = (1 << 23);
-  static const uint64_t c_ResponseGeneration       = (1 << 24);
+  static const uint64_t c_TACcal                   = (1 << 5);  // = 32
+  static const uint64_t c_TACcut                   = (1 << 6);  // = 64
```
fhagemann: Just checking: Is this still correct, both still being within the same module instead of different modules?

src/MModuleTACcal.cxx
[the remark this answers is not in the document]
ckierans: I like the name TACCalibration vs TACCut.

include/MAssembly.h
NicoleRodriguezCavero: It is not, I correcting this as well as the name.

src/MGUIExpoPlotSpectrum.cxx
NicoleRodriguezCavero: These tabs were relabeled because they show the TAC distribution and energy spectra after TAC cuts (not calibrations) are applied

Image

src/MSubModuleDepthReadout.cxx
```diff
@@ -98,11 +98,11 @@ bool MSubModuleDepthReadout::Initialize()
   }
 
   // Load TAC calibration parameters
-  MModuleTACcut TACcut;
-  TACcut.SetTACCalFileName(m_TACCalFileName);
-  if (TACcut.LoadTACCalFile(m_TACCalFileName) == true) {
+  MModuleTACCalibration TACcal;
```
ckierans: Nitpicky, but maybe use TACCal vs TACcal as the MModuleTACCalibration instance to be consistent with the capitalization of the "C" in Cal

src/MModuleTACCalibration.cxx
```diff
@@ -0,0 +1,498 @@
...
+  // Set all modules, which have to be done before this module
+  AddPreceedingModuleType(MAssembly::c_EventLoader);
+  AddPreceedingModuleType(MAssembly::c_EnergyCalibration);
+
+  // Set all types this modules handles
+  AddModuleType(MAssembly::c_TACCalibration);
+
+  // Set all modules, which can follow this module
+  AddSucceedingModuleType(MAssembly::c_StripPairing);
+  AddSucceedingModuleType(MAssembly::c_DepthCorrection);
```
ckierans: The DepthCorrection module isn't actually allowable here since it has it's own dependance that StripPairing must be performed first, so I think this can be removed.

src/MModuleTACCalibration.cxx
```diff
@@ -0,0 +1,498 @@
...
+
+  // Set the module name --- has to be unique
+  m_Name = "TAC Calibration";
+
+  // Set the XML tag --- has to be unique --- no spaces allowed
+  m_XmlTag = "XmlTagTACCalibration";
+
+  // Set all modules, which have to be done before this module
+  AddPreceedingModuleType(MAssembly::c_EventLoader);
+  AddPreceedingModuleType(MAssembly::c_EnergyCalibration);
```
ckierans: So now we're enforcing the EnergyCalibration to happen before the TAC Calibration? This is different than what we previously had where we left it up to the user (I think I've always done TAC first then Energy), which is probably the wrong approach! Have we confirmed there is not change of behavior with TAC Calibration -> Energy Calibration vs Energy Calibration -> TAC Calibration? I don't think it should necessarily matter, but I want to make sure we're being intentional about the order.

Honestly, this got me looking at the other Nuclearizer dependancies, and I do think we should enforce only one option for the order for TAC/Energy, so I'm happy to see this change!

src/MModuleTACCalibration.cxx
```diff
@@ -0,0 +1,498 @@
...
+      }
+
+      // Need at least slope and offset
+      if (m_TACCal[DetID][SideIndex][StripID].size() < 2) {
+        cout<<m_XmlTag
+            <<": Error: StripID "<<StripID
+            <<" on side "<<Side
+            <<" does not have enough TAC calibration parameters - skipping event"
+            <<endl;
+        return false;
```
ckierans: Do you mean to use `return false` or `continue` here? Return false will stop the whole Nuclearizer run if one of these conditions are met, it won't just skip the event. `continue` will just skip the event currently being processed. See for example lines 168 -- 174 in the MModuleEnergyCalibration for when there is no valid energy calibration for a striphit. If we want to not throw a full error and stop Nuclearizer if a TAC Calibration for a single strip is unknown, then we probably should introduce a TACCalibrationError flag, similar to the EnergyCalibrationError flag, and downgrade this to a Warning vs an Error. 

I'm also wondering if it makes more sense to combine the valid calibration parameters + at least two parameters describing the calibration into one warning/error.

Take a closer look at these error messages and think about when we really need to stop Nuclearizer, and when we just need a flag for an event.

src/MModuleTACCalibration.cxx
```diff
@@ -0,0 +1,498 @@
...
+        if ((SHTiming < TotalOffset) || (SHTiming < MaxTAC - m_CoincidenceWindow)) {
+          Passed = false;
+        }
+      }
+    }
+
+    if (Passed == true) {
+      ++i;
+    } else {
+      Event->RemoveStripHit(i);
```
ckierans: Is there any reason to track the events for which we've removed striphits with the TAC cut? Does this give any indication of the quality of the event, or any information about the rest of the striphits?

ckierans (commented):
Nice rework, @NicoleRodriguezCavero! I just have a few questions about the error messages, and the ordering of the modules.

Will this PR close either Issue #64 or #63 ?

src/MSubModuleDepthReadout.cxx
fhagemann: Probably best to call this variable `TACCalibration` ;)

src/MModuleTACCalibration.cxx
NicoleRodriguezCavero: I am moving this discussion to issue #207 because (1) we might need further tests with NN and module orders and (2) there will be upcoming analysis work that might want to use these modules independently. I think this should be done independently from this PR.

src/MModuleTACCalibration.cxx
NicoleRodriguezCavero: I don't think we need to track events where strip hits are removed by the TAC cuts. These cuts are removing hits associated with the ASIC timing bug or chance-coincident hits which should not affect the quality of the remaining hits in the event.

NicoleRodriguezCavero:
Addressing @ckierans comments (thank you!!)

> Will this PR close either Issue #64 or #63 ?

This PR will resolve Issue #63, we have established that is the cut that we want. It should not resolve Issue #64 since that is more of an analysis/simulation task.

> So now we're enforcing the EnergyCalibration to happen before the TAC Calibration?

I created Issue #207 to address this concern separately from the PR.

> Do you mean to use return false or continue here? Return false will stop the whole Nuclearizer run if one of these conditions are met, it won't just skip the event...

I have updated the function following the `EnergyCalibration` conventions. Missing/invalid TAC calibrations for an individual strip now gives a `TACCalibrationError` flag on the event and continues instead of returning `false`.

Fatal errors are now only applied when the calibration file or module cannot be initialized correctly.

I also combined the "missing calibration entry" and "fewer than two calibration parameters" as Carolyn suggested.

include/MModuleTACCalibration.h
```diff
@@ -9,90 +9,88 @@
...
 // MEGAlib libs:
 #include "MGlobal.h"
 #include "MModule.h"
-#include "MGUIExpoTACcut.h"
-#include "MGUIExpoPlotSpectrum.h"
 
 
 // Forward declarations:
+class MGUIExpoTACcut;
+class MGUIExpoPlotSpectrum;
```
fhagemann: Why did you change this from `#include` to a forward declaration?

src/MGUIExpoPlotSpectrum.cxx
```diff
@@ -62,7 +62,11 @@ MGUIExpoPlotSpectrum::MGUIExpoPlotSpectrum(MModule* Module) : MGUIExpo(Module)
 
   // Set the new title of the tab here:
   if (Module != nullptr) {
+      if (Module->GetName() == "TAC Calibration") {
+        m_TabTitle = "Energy Spectrum (TAC cuts)";
+      } else {
       m_TabTitle = "Energy Spectrum (" + Module->GetName() + ")";
```
fhagemann:
```suggestion
        m_TabTitle = "Energy Spectrum (" + Module->GetName() + ")";
```

include/MGUIOptionsTACCalibration.h
```diff
@@ -43,28 +43,37 @@ class MGUIEEntry;
...
   //! Actions after the Apply or OK button has been pressed
   virtual bool OnApply();
 
+  //! Toggle TAC cut radio buttons and coincidence window entry
+  void ToggleRadioButtons(int WidgetID);
 
+  //! Widget IDs
+  enum {
+  c_TACCutIgnore = 120,
+  c_TACCutApply = 121
```
fhagemann:
```suggestion
    c_TACCutIgnore = 120,
    c_TACCutApply = 121
```

fhagemann (approved):
Looks good!
Very minor comments from my side only

include/MModuleTACCalibration.h
NicoleRodriguezCavero: The `#include` are now called only in`src/MModuleTACCalibration.cxx` where we are using the methods of `MGUIExpoTACcut` and `MGUIEXpoPlotSpectrum`. In the header file the forward declaration is enough.

ckierans: Whoops, accidentally unresolved the convo when reading over things. I resolved it again!

ckierans (approved):
Thanks for the changes, Nicole! Things look good.

PR #197, opened 2026-08-18
CHG: Update XML tags of `EnergyCalibration` and `DepthCalibration`

### LINKED ISSUE #193: Consistent use of XML tags

fhagemann:
Addressing #193:
The renaming of the TAC calibration module in #194 might require folks to update their nuclearizer configuration files, so I would suggest to also tackle updating the XML tags of `EnergyCalibration` and `DepthCalibration` to `XmlTagEnergyCalibration` and `XmlTagDepthCalibration` in the same run, so that the config files only have to be updated once.

I also added a small note to the `CodingConventions.md` that XML tags of `MModules` should start with `XmlTag`.

### DIFF
--- a/CodingConventions.md
+++ b/CodingConventions.md
@@ -37,14 +37,14 @@ To format a file do the following:
 
 ## New classes
 
-For new classes, copy and modify an existing one, or use the MModuleTemplate class as tempplate.
+For new classes, copy and modify an existing one, or use the MModuleTemplate class as template.
 
 ## Naming
 
 ### Classes
 
 * Due to historic reasons, all classes start with M.
-* Modules start with MModule
+* Modules start with MModule, with XML tags starting with XmlTag
 * Option GUI's start with MGUIOptions
 * Expo (= data display) GUI's start with MGUIExpo
 
--- a/Config/Nuclearizer_EM_DEE.cfg
+++ b/Config/Nuclearizer_EM_DEE.cfg
@@ -73,23 +73,16 @@
       <MinimumStripPairingReducedChiSquare>-1</MinimumStripPairingReducedChiSquare>
       <MaximumStripPairingReducedChiSquare>inf</MaximumStripPairingReducedChiSquare>
     </XmlTagEventFilter>
-    <EnergyCalibration>
+    <XmlTagEnergyCalibration>
       <FileName />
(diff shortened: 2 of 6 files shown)


fhagemann:
Feel free to approve / request changes, but I would propose to wait with merging this until #194 is approved, and then merge both PRs at the same time.

ckierans (approved):
I love the consistency :)

PR #200, opened 2026-08-19
Added flag for events with an ADC value higher than 14000.

### LINKED ISSUE #40: Handle non-linear gain/ADC overflow
The gain in the ASIC becomes non-linear before the max ADC/overflow bin is reached. This is not calibrated, so the easiest thing will be to implement a cut in ADC ~14000 (estimated from Clio and Brent May 2025) and flag these hits as overflow.

nlopez-code:
Included in unit test as well
( #40 )

### DIFF
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -139,11 +139,6 @@ class MStripHit
   //! Return whether the strip has triggered (ADC values above slow threshold)
   bool HasTriggered() const { return m_HasTriggered; }
 
-  //! Set whether the strip has a high ADC value (close to the ADC saturation limit)
-  void HasHighADC(bool HighADC) { m_HasHighADC = HighADC; }
-  //! Return whether the strip has a high ADC value (close to the ADC saturation limit)
-  bool HasHighADC() const { return m_HasHighADC; }
-
   //! TODO: Rename to HasTiming()
   //! Set the calibrated-timing flag
   void HasCalibratedTiming(bool CalibratedTiming) { m_HasCalibratedTiming = CalibratedTiming; }
@@ -216,8 +211,6 @@ class MStripHit
   bool m_HasFastTiming;
   //! True if the hit has calibrated timing
   bool m_HasCalibratedTiming;
-  //! True if the ADC value is close to the ADC saturation limit
-  bool m_HasHighADC;
 
   //! Origin interaction IDs from the simulation
   vector<int> m_Origins;
--- a/include/MStripHit.h
+++ b/include/MStripHit.h
@@ -139,6 +139,11 @@ class MStripHit
   //! Return whether the strip has triggered (ADC values above slow threshold)
   bool HasTriggered() const { return m_HasTriggered; }
 
+  //! Set whether the strip has a high ADC value (close to the ADC saturation limit)
+  void HasHighADC(bool HighADC) { m_HasHighADC = HighADC; }
+  //! Return whether the strip has a high ADC value (close to the ADC saturation limit)
+  bool HasHighADC() const { return m_HasHighADC; }
+
   //! TODO: Rename to HasTiming()
   //! Set the calibrated-timing flag
   void HasCalibratedTiming(bool CalibratedTiming) { m_HasCalibratedTiming = CalibratedTiming; }
@@ -211,6 +216,8 @@ class MStripHit
   bool m_HasFastTiming;
   //! True if the hit has calibrated timing
   bool m_HasCalibratedTiming;
+  //! True if the ADC value is close to the ADC saturation limit
+  bool m_HasHighADC;
 
   //! Origin interaction IDs from the simulation
   vector<int> m_Origins;
--- a/src/MModuleEnergyCalibration.cxx
+++ b/src/MModuleEnergyCalibration.cxx
@@ -2,14 +2,12 @@
  * MModuleEnergyCalibration.cxx
  *
  *
- * Copyright (C) by Andreas Zoglauer, Mark Bandstra, Carolyn Kierans,
- * Jackie Beechert, Robin Anthony-Peterson.
+ * Copyright (C) 2008-2008 by Andreas Zoglauer.
  * All rights reserved.
  *
  *
  * This code implementation is the intellectual property of
- * Andreas Zoglauer, Mark Bandstra, Carolyn Kierans, 
- * Jackie Beechert, Robin Anthony-Peterson
+ * Andreas Zoglauer.
  *
  * By copying, distributing or modifying the Program (or any work
  * based on the Program) you indicate your acceptance of this statement,
@@ -24,29 +22,16 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-// Standard libs:
-#include <fstream>
-#include <iostream>
-#include <map>
-using namespace std;
 
 // Include the header:
 #include "MModuleEnergyCalibration.h"
 
+// Standard libs:
 
 // ROOT libs:
 #include "TGClient.h"
-#include "TFile.h"
 
 // MEGAlib libs:
-#include "MString.h"
-#include "MStreams.h"
-
-// Nuclearizer libs:
-#include "MReadOutElement.h"
-#include "MReadOutElementDoubleStrip.h"
-#include "MGUIOptionsEnergyCalibration.h"
-#include "MGUIExpoPlotSpectrum.h"
 
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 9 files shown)


fhagemann (approved):
This is one of the most beautifully written PRs I've ever seen.
I have one tiny comment on your code, and will follow with instructions how to get the unit tests to pass :)

fhagemann:
If you look at the test logs, you can see the following:
```
PASS UTNDEEStripHit (Passed tests: 40, Failed tests: 0)
UTNEndToEnd_406-1 (1/7 failed)
UTNEndToEnd_542-1 (1/7 failed)
PASS UTNHit (Passed tests: 139, Failed tests: 0)
PASS UTNModuleLoaderMeasurementsHDF (Passed tests: 118, Failed tests: 0)
PASS UTNReadOutAssembly (Passed tests: 255, Failed tests: 0)
PASS UTNStripHit (Passed tests: 145, Failed tests: 0)
PASS UTNStripMap (Passed tests: 236, Failed tests: 0)
```
which states that something in the end-to-end tests for detectors 406-1 and 542-1 failed.

You can run those tests on your local machine by running
```
make -j16
make -j16 unittests
UTNEndToEnd_406-1
```
Which in my case yields the following output:
```
FAILED: End-to-end test 406-1  <-- output file
   Description: The generated output is numerically equivalent to the reference
   Expected:    numerically equivalent files
   Output:      
      line 1340:
        expected:  QA StripPairing (Best reduced chi square is not below 25 (466.978427)) (GR Hit: Detector ID 0 and Energy 294.916417)
        test:      QA HighADC (High ADC value 15804 for Detector: 0, side: LV, strip: 20) (High ADC value 16013 for Detector: 0, side: LV, strip: 21) (High ADC value 16006 for Detector: 0, side: LV, strip: 22) (High ADC value 15449 for Detector: 0, side: LV, strip: 23) (High ADC value 15058 for Detector: 0, side: HV, strip: 53) (High ADC value 16015 for Detector: 0, side: HV, strip: 54) (High ADC value 16038 for Detector: 0, side: HV, strip: 55)

Unit test: UTNEndToEnd_406-1
Passed tests: 7
Failed tests: 1
```
Essentially this means that, if we run `nuclearizer` using the updated code in this PR, we would get a different output (which makes sense here: we're now producing additional quality flags, and they will end up appearing in the final `.tra` file.

Some background information:
In the unit tests, we automatically run `nuclearizer` using the config files in `resource/unittestdata/406-1` (or `542-1` respectively) via `nuclearizer -c hdf5-to-tra.nuclearizer.cfg` and compare the output to `hdf5-to-tra.reference.tra`.
You're now getting the (expected) error that the newly generated `.tra` file does not match the reference `.tra` file.
To resolve this, you might need to update those reference files, such that they include the quality flags that you are adding in this PR. You can achieve this by manually running `nuclearizer -c hdf5-to-tra.nuclearizer.cfg` in both the `406-1` and `542-1` directories yourself, renaming the created output file `hdf5-to-tra.tra` to `hdf5-to-tra.reference.tra` to update the reference file, and pushing the updated reference file to this PR.

Let me know if you run into any issues tackling this, always happy to help! :)

src/MModuleEnergyCalibration.cxx
```diff
@@ -162,6 +162,16 @@ bool MModuleEnergyCalibration::AnalyzeEvent(MReadOutAssembly* Event)
...
     MReadOutElementDoubleStrip R = *dynamic_cast<MReadOutElementDoubleStrip*>(SH->GetReadOutElement());
 
+    // Flag strip hits whose ADC value is close to the ADC saturation limit, since their
+    // calibrated energy is not trustworthy. The hit is kept, it is only marked.
+    if (SH->GetADCUnits() > m_HighADCThreshold) {
+      SH->HasHighADC(true);
+      if (g_Verbosity >= c_Warning) {
+        cout << m_XmlTag << ": Warning: High ADC value " << SH->GetADCUnits() << " for read-out element " << R << endl;
+      }
+      Event->SetHighADC_QualityFlag("High ADC value " + to_string(SH->GetADCUnits()) + " for " + R.ToString().Data());
```
fhagemann: Tiny comment when composing the QualityFlag Text: you might want to convert the `ADC` value to an (unsigned) `int` to avoid a lot of unnecessary zeros in the message:
```suggestion
      if (g_Verbosity >= c_Warning) {
        cout << m_XmlTag << ": Warning: High ADC value " << (int) SH->GetADCUnits() << " for read-out element " << R << endl;
      }
      Event->SetHighADC_QualityFlag("High ADC value " + to_string((int) SH->GetADCUnits()) + " for " + R.ToString().Data());
```

This will be resolved by itself once we tackle #176, but we can hack this here for now ;)

Before:
```
QA HighADC (High ADC value 15804.000000 for Detector: 0, side: LV, strip: 20) (High ADC value 16013.000000 for Detector: 0, side: LV, strip: 21) (High ADC value 16006.000000 for Detector: 0, side: LV, strip: 22) (High ADC value 15449.000000 for Detector: 0, side: LV, strip: 23) (High ADC value 15058.000000 for Detector: 0, side: HV, strip: 53) (High ADC value 16015.000000 for Detector: 0, side: HV, strip: 54) (High ADC value 16038.000000 for Detector: 0, side: HV, strip: 55)
```

After:
```
QA HighADC (High ADC value 15804 for Detector: 0, side: LV, strip: 20) (High ADC value 16013 for Detector: 0, side: LV, strip: 21) (High ADC value 16006 for Detector: 0, side: LV, strip: 22) (High ADC value 15449 for Detector: 0, side: LV, strip: 23) (High ADC value 15058 for Detector: 0, side: HV, strip: 53) (High ADC value 16015 for Detector: 0, side: HV, strip: 54) (High ADC value 16038 for Detector: 0, side: HV, strip: 55)
```

fhagemann (commented):
Also, looks like my comment didn't go through ^^

include/MStripHit.h
```diff
@@ -139,6 +139,11 @@ class MStripHit
   //! Return whether the strip has triggered (ADC values above slow threshold)
   bool HasTriggered() const { return m_HasTriggered; }
 
+  //! Set whether the strip has a high ADC value (close to the ADC saturation limit)
+  void HasHighADC(bool HighADC) { m_HasHighADC = HighADC; }
+  //! Return whether the strip has a high ADC value (close to the ADC saturation limit)
+  bool HasHighADC() const { return m_HasHighADC; }
```
ckierans: I'm wondering if it's appropriate to be defining a StripHit flag for HighADC? 

StripHits with the > HighADC value should be treated with caution, but I think the HighADC quality flag for an event, as you have defined it in MReadoutAssembly, might be enough?

The flags here for StripHits, such as HasFastTiming, IsGuardRing, etc, are derived from the raw data and describe the *type* of strip hit but not necessarily the *quality* of the strip hit. I think the HighADC flag is most similar to the SlowThresholdCut (other then us not removing the strip hit), so we should more closely follow that example. And if that's the case, then I think you did a lot more work then you needed to.

src/MStripHit.cxx
```diff
@@ -227,21 +228,25 @@ void MStripHit::StreamRoa(ostream& S, bool WithADC, bool WithTAC, bool WithEnerg
...
 {
   // Return the bitwise strip-hit flags
-  // Currently, 3 bits:
-  //   v = Has fast timing
-  //    v = Is a nearest neighbor strip hit
-  //     v = Is a guard ring strip hit
-  // 0b111u
-
-  unsigned int Flags = 0b000u;
+  // Currently, 4 bits:
```
ckierans: This gives me pause. I don't think we want HighADC to be looped in with these higher-level flags.

ckierans (commented):
Nice work on your first Nuclearizer PR, @nlopez-code !!!

That said, I think you overdid the flags a bit :). The QualityFlag in MReadOutAssembly is probably sufficient, where as the flag in the MStripHit class is probably too much, at least in my interpretation of the purpose of those flags.

We also need to decide if we should do anything with the ADC values if they are measured >14000. Should we treat 14000 ADC truly like a software overflow bin and reassign the high ADC values to be 14000, or do we just keep it as is and treat these events with caution to be potentially filtered out in revan/mimrec? This probably needs to be investigated further with some high energy data to determine the best approach, but that of course can be a TODO.

Perhaps for now just add a comment in the code when you raise the High ADC warning that we might want to revisit how we handle the high ADC values.

fhagemann:
Is this missing the updated `MModuleEnergyCalibration.cxx`?
In the current version of the files, we are not setting the flag anywhere (?)

src/MModuleEnergyCalibration.cxx
```diff
@@ -112,411 +86,9 @@ MModuleEnergyCalibration::~MModuleEnergyCalibration()
...
-
-    // Flag strip hits whose ADC value is close to the ADC saturation limit, since their
-    // calibrated energy is not trustworthy. The hit is kept, it is only marked.
-    if (SH->GetADCUnits() > m_HighADCThreshold) {
-      SH->HasHighADC(true);
-      if (g_Verbosity >= c_Warning) {
-        cout << m_XmlTag << ": Warning: High ADC value " << SH->GetADCUnits() << " for read-out element " << R << endl;
-      }
-      Event->SetHighADC_QualityFlag("High ADC value " + to_string(SH->GetADCUnits()) + " for " + R.ToString().Data());
-    }
```
fhagemann: Seems to me like we still would like to keep this (or any other related) code :)

src/MModuleEnergyCalibration.cxx
```diff
@@ -162,6 +162,16 @@ bool MModuleEnergyCalibration::AnalyzeEvent(MReadOutAssembly* Event)
     MStripHit* SH = Event->GetStripHit(i);
     MReadOutElementDoubleStrip R = *dynamic_cast<MReadOutElementDoubleStrip*>(SH->GetReadOutElement());
 
+    // Flag strip hits whose ADC value is close to the ADC saturation limit, since their
+    // calibrated energy is not trustworthy. The hit is kept, it is only marked.
+    if (SH->GetADCUnits() > m_HighADCThreshold) {
+      SH->HasHighADC(true);
+      if (g_Verbosity >= c_Warning) {
+        cout << m_XmlTag << ": Warning: High ADC value " << SH->GetADCUnits() << " for read-out element " << R << endl;
```
fhagemann: You might also want to add the `(int)` in this line:
```suggestion
        cout << m_XmlTag << ": Warning: High ADC value " << (int) SH->GetADCUnits() << " for read-out element " << R << endl;
```

src/MModuleEnergyCalibration.cxx
```diff
@@ -162,6 +162,16 @@ bool MModuleEnergyCalibration::AnalyzeEvent(MReadOutAssembly* Event)
     MStripHit* SH = Event->GetStripHit(i);
     MReadOutElementDoubleStrip R = *dynamic_cast<MReadOutElementDoubleStrip*>(SH->GetReadOutElement());
 
+    // Flag strip hits whose ADC value is close to the ADC saturation limit, since their
+    // calibrated energy is not trustworthy. The hit is kept, it is only marked.
+    if (SH->GetADCUnits() > m_HighADCThreshold) {
+      SH->HasHighADC(true);
```
fhagemann: ~This should be `Event->HasHighADC(true);`?~

fhagemann: Edit: Actually, this line can completely be removed (the StripHit has no `m_HasHighADC` anymore), as `Event->SetHighADC_QualityFlag` later in the code takes care of that 

```suggestion
```

fhagemann (approved):
Unit tests pass, looks good to me! :)

fhagemann:
While I love the commit messages, we might want to squash them when merging ^^
image

fhagemann:
@ckierans Thanks for pushing the last commit. I can merge this once I get your blessing

ckierans:
I got sidetracked while waiting for the checks to pass. I'll squash and merge now, as you suggested.

Issue #198, opened 2026-08-19
Clean up EventSaver module order dependancies


ckierans:
After selecting the strip pairing module in Nuclearizer, the allowable modules are shown below:

<img width="363" height="340" alt="Image" src="https://github.com/user-attachments/assets/4f0e1da0-fa33-4327-b42f-d3d0142bbe8e" />

This means that someone can select saving an event in L1b/L2 format without running the DepthCalibration module. This needs to be changed in the MModuleSaverMeasurementsFITS.cxx file:
```
153 void MModuleSaverMeasurementsFITS::ConfigurePreceedingModules()
154 {
155   m_PreceedingModules.clear();
156   m_PreceedingModulesHardRequirement.clear();
157 
158   AddPreceedingModuleType(MAssembly::c_EventLoader);
159   if (m_OutputDataLevel != 0) {
160     AddPreceedingModuleType(MAssembly::c_StripPairing);
161   }
162 }
```
Also, there is an GUI option "Save events to FITS files (L1b/L2)" and then individual options to "Save events to L1a FITS" and "Save events to L1b FITS" etc, so this is redundant and the XmlTags need to be cleaned up.

As a reminder, this is the data level definition:
- *L1a* contains the raw telemetered data (a) event time, event ID, event type, number of interactions, detector number, detector side, strip number, pulse height ADC, and arrival time TAC. 
- *L1b* contains data after calibration and event reconstruction
- *L2* contains a cleaned version of the event data from Level 1b

The dependancies should be as follows (I think!):
- Save events to L0 binary (DD packets): after either cosima simulation loader +  DEE or any measurement loaders, but before any calibration modules.
- Save events to L1a FITS: after either cosima simulation loader +  DEE or any measurement loaders, but before any calibration modules.
- Save events to L1b FITS: only after depth calibration. But I think we want this option both before or after Event reconstruction (revan)? The only thing that is modified after reconstruction is the sequence of hit information, so that can be empty in the L1b file if the reconstruction hasn't been performed, I assume?
- Save events to L2 FITS: only after Event reconstruction (revan)

mwy-wing:
For the L1b, I think it makes more sense to have it available only after event reconstruction, since the definition of L1b is data after calibration and event reconstruction.

ckierans:
Closed with PR #201

PR #201, opened 2026-08-20
FIX: Clean up the module ordering for the FITS and L0 savers

mwy-wing:
This PR addresses issues from #198 . I made a few changes to how the module picking behaves:

1. The generic MModuleSaverMeasurementsFITS is removed, since it was redundant. I also removed the m_OutputDataLevelCombo, because it is no longer needed.
2. The L0 saver is now available only after an event loader and the DEE.
3. The L1a saver now needs to immediately follow a measurement loader, so it does not appear after the calibration steps.
4. The L1b and L2 savers come after strip pairing, depth calibration, and event reconstruction.

### DIFF
--- a/src/MModuleSaverMeasurementsL0.cxx
+++ b/src/MModuleSaverMeasurementsL0.cxx
@@ -67,8 +67,9 @@ MModuleSaverMeasurementsL0::MModuleSaverMeasurementsL0() : MModule()
 
   // Set all modules, which have to be done before this module:
   // The L0 saver writes simulated events into a binary file - requires an event loader and the detector effects engine module before it
+  // The DEE must be immediately before it
   AddPreceedingModuleType(MAssembly::c_EventLoader);
-  AddPreceedingModuleType(MAssembly::c_DetectorEffectsEngine);
+  AddPreceedingModuleType(MAssembly::c_DetectorEffectsEngine, true, true);
 
   // Set all types this modules handles
   AddModuleType(MAssembly::c_EventSaver);
--- a/src/MModuleSaverMeasurementsL0.cxx
+++ b/src/MModuleSaverMeasurementsL0.cxx
@@ -65,11 +65,16 @@ MModuleSaverMeasurementsL0::MModuleSaverMeasurementsL0() : MModule()
   // Set the XML tag --- has to be unique --- no spaces allowed
   m_XmlTag = "XmlTagSaverMeasurementsL0";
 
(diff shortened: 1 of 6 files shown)


mwy-wing:
The tests failed because ImmediatelyPreceeding and SetTypeExclusive are new features I added to MEGAlib.
We should only merge this PR after zoglauer/megalib#127.

zoglauer:
Mark it as draft - I take a look at the MEGAlib PR either tonight or tomorrow

zoglauer:
Yes, I think the L0 saver was exactly designed for case 1.

fhagemann:
I was able to run the DEE after updating megalib to `feature/supervisor-hardening` and using this PR.
However, removing `m_PreceedingModulesHardRequirement` in `feature/supervisor-hardening` resulted in some errors when switching to other branches in nuclearizer, without downgrading megalib to the current `develop-cosi`.

If we want to merge this, I would suggest to merge this PR and `feature/supervisor-handling` at the same time, and require everyone to update megalib and to rebase their branches on top of this PR.

src/MModuleSaverMeasurementsL0.cxx
```diff
@@ -65,12 +65,17 @@ MModuleSaverMeasurementsL0::MModuleSaverMeasurementsL0() : MModule()
   // Set the XML tag --- has to be unique --- no spaces allowed
   m_XmlTag = "XmlTagSaverMeasurementsL0";
 
-  // Set all modules, which have to be done before this module
+  // Set all modules, which have to be done before this module:
+  // The L0 saver writes simulated events into a binary file - requires an event loader and the detector effects engine module before it
   AddPreceedingModuleType(MAssembly::c_EventLoader);
+  AddPreceedingModuleType(MAssembly::c_DetectorEffectsEngine);
```
fhagemann: As noted earlier, maybe force the L0 saver to only be selectable right after the DEE (to only allow for what I called "Case 1")

fhagemann:
Whatever changes we have to make after merging the changes in `megalib/feature/supervisor-hardening` should be added to BOTH `nuclearizer/develop/em`, but also `nuclearizer/main`.
(I think we had this in the past, where updating `nuclearizer/develop/em` but not `nuclearizer/main` led to people installing "fresh" megalib/nuclearizer using cosi-setup to have compilation errors)

zoglauer (approved):

PR #202, opened 2026-08-21
Removed HasTriggered variable from MStripHit

ckierans:
As agreed upon in the August 18th CDEE meeting, we're removing the HasTriggered variable from MStripHit. This variable wasn't used in the HDF measurement loader, and it is redundant (but opposite to) IsNearestNeighbor. 

Note that I kept the HasTriggered variable as part of the MShieldCrystalHit and MCrystalHit class (it seems like one of these is a duplicate...)

### DIFF
--- a/include/MDEEStripHit.h
+++ b/include/MDEEStripHit.h
@@ -82,6 +82,8 @@ struct MDEEStripHit
   double m_Energy;
   //! The measured ADC value
   unsigned int m_ADC;
+  //! If the strip was a nearest neighbor
+  bool m_IsNearestNeighbor;
   //! If the strip has fast timing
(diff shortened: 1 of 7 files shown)


fhagemann (commented):
~I made adjustments to this PR to get the unit level tests to pass again.~

~However, we might want to think, that instead of completely removing the information on `m_HasTriggered` to an output file, we might want to replace it with the information in `m_IsNearestNeighbor` instead.~

fhagemann (commented):
I force-pushed making this PR non-breaking when it comes to reading/writing to DAT/ROA files (by replacing `m_HasTriggered` streaming with (the inverse of) `m_IsNearestNeighbor`.

PR #203, opened 2026-08-21
ADD: calibration files from unit-level testing to `resource/calibration`

fhagemann:
In this PR, we add the full ecal, taccal, depth splines and depth coefficients for the 16 flight detectors as determined from the unit-level calibrations, assuming the thicknesses from the table below ("E (sims)") and realized in https://github.com/cositools/massmodel-cosi-payload/pull/3:
image

Also added are the combined files for all 16 detectors, derived from the individual unit-level files.

- [x] I still need to run a final cross-check to make sure that all the files here are the FINAL calibration file as reported in the Unit Level Reports, which will require some extra time. 
- [x] I also still need to do a final cross-check to see if the DEE / forward pipeline accepts them based on the thicknesses updated in https://github.com/cositools/massmodel-cosi-payload/pull/3.

Note that for PAYLOAD calibrations, we will use updated thickness (and temperature) values (see https://github.com/cositools/massmodel-cosi-payload/pull/4), that will be added to `resource/calibration/COSISMEX/payload` (or whatever we might call the repository) once they become available.


fhagemann:
I had one more thorough round of cross-checking the calibration files in this PR against:
1. what's reported in the unit-level reports
2. what's reported in the ARM summary slides by Sean/Julian

image

Some notes:
- Marked in $\color{red}\textsf{red}$ is where the files deviate from what is reported in the unit-level reports.
  For HP52406-3, this is intentional: we took three TAC calibrations with timestamp `20250821`, `20250826` and `20250829`. It looks like the `20250821` was different from the `20250826`/`20250829` one:
  Screenshot from 2026-08-24 10-43-25
  
  As the depth coefficients file was generated with `20250821`, I chose to keep the `20250821` TAC calibration file in this PR instead of one of the "newer" `20250826`/`20250829` ones. If we want to keep a "newer"  `20250826`/`20250829` TAC calibration file, we would need to re-run the depth calibration for this detector.

- Marked in $\color{blue}\textsf{blue}$ are the values where another file was used for the ARM calibration plots.
  - For HP52358-2, HP52432-1, and HP52358-1, the ARM plots were composed based on an energy calibration that did not apply TAC cuts. The energy calibrations were updated in September to include TAC cuts, and are also the ones reported in the unit-level reports.
  - For HP52542-2, HP52407-3, and HP52406-3, the ARM plots were composed using the quick ecal instead of the full ecal. Here, I'm using the energy calibration files reported in the unit-level reports.

One final thing to note: seems like for HP52358-3, the full energy calibration was not re-run to include TAC cuts.

fhagemann:
I also seem to be able to run the DEE with all those files and the mass model in https://github.com/cositools/massmodel-cosi-payload/pull/3

ckierans (approved):
This is a huge effort, @fhagemann !!! Thank you so much for the careful book keeping!

